PR #3, opened 2022-05-03
Add basic nest skeleton. 

kailamjeter:
Includes tasks 1-3 of https://app.shortcut.com/blueconduit/story/3679/initialize-api. 

This is the result of running 'nest new api' - I did not otherwise modify the files in api/.

Added all files included at https://github.com/BlueConduit/tributary/tree/main/services except the .gitignore. Added relevant .gitignore rules to the global .gitignore.

### DIFF
--- a/api/src/app.controller.spec.ts
+++ b/api/src/app.controller.spec.ts
@@ -1,0 +1,22 @@
+import { Test, TestingModule } from '@nestjs/testing';
+import { AppController } from './app.controller';
+import { AppService } from './app.service';
+
+describe('AppController', () => {
+  let appController: AppController;
+
+  beforeEach(async () => {
+    const app: TestingModule = await Test.createTestingModule({
+      controllers: [AppController],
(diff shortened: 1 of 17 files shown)


api/src/app.controller.spec.ts
```diff
@@ -0,0 +1,22 @@
+import { Test, TestingModule } from '@nestjs/testing';
```
beekley, line 1: Could you add a brief comment describing the purpose of this and other `ts` files here? I can even just be descriptions from https://docs.nestjs.com/first-steps#setup. Since this is new for all of us, I'd like to err on the side of over-documentation.

beekley (approved):
I don't fully understand the NestJS stuff, but I'm reading through the docs now.

api/src/app.controller.spec.ts
kailamjeter: Done!

PR #5, opened 2022-05-04
Add vue router for / and /about routes. Links to the same page for now

justinebreuch:
Adds a header for the links / and /about. We need to request a vertically centered version of the logo so that we don't have to add hacky css to vertically center navbar items


client/src/assets/styles/global.scss
```diff
@@ -0,0 +1,5 @@
+// Typography
```
kailamjeter, line 1: Will this file contain all styles for the app? Saw something about importing styles in App.vue so curious as to how this will be extended for other components?

client/src/router.js
```diff
@@ -0,0 +1,16 @@
+import { createRouter, createWebHistory } from 'vue-router'
+import MapContainer from './components/MapContainer.vue'
+
+export default createRouter({
```
kailamjeter, line 4: Can we add a comment here just like /** Defines all routes for app. **/

kailamjeter (approved):

client/src/router.js
justinebreuch: Good note thanks

client/src/assets/styles/global.scss
kailamjeter: Makes sense! We should leave a comment on this file to that effect.

PR #8, opened 2022-05-12
Dummy data

kailamjeter:
## Description

Addresses: [The map frontend displays a shape that is stored in the DB](https://app.shortcut.com/blueconduit/story/4759/the-map-frontend-displays-a-shape-that-is-stored-in-the-db?cf_workflow=500000067&ct_workflow=all&vc_group_by=day)

Calls API to get s3 bucket URL of water boundary geojson file and displays this in the FE. This does not yet involve the DB.

### New

Axios dependencies [link to documentation](https://axios-http.com/docs/intro). We can discuss whether this is the way we want to keep going but seems like this + native API support is all and this seems more robust.

### Changed

Source of data from static s3 bucket URL --> an API call. Additionally fixes map styling so that data renders. 

## Testing and Reviewing

Run npm run serve to view changes. Warning that the data takes a few seconds to render.

## Screenshot

Screen Shot 2022-05-12 at 4 49 44 PM

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -9,6 +9,7 @@
       "version": "0.1.0",
       "dependencies": {
         "@blueconduit/copper": "^1.1.0",
+        "axios": "^0.27.2",
         "core-js": "^3.8.3",
         "mapbox-gl": "^2.8.2",
         "vue": "^3.2.13",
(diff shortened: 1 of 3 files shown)


beekley:
Can you link to a screenshot of what the map looks like if we ran it?

beekley (approved):

PR #11, opened 2022-05-18
Add presence test for CDK

beekley:
## Description

Addresses: [set-up-the-map-data-db](https://app.shortcut.com/blueconduit/story/3718/set-up-the-map-data-db)

Adds a basic test that only checks for presence of resources. I don't think it's particularly useful to add a bunch of tests here early on in the project -- we can add more tests as we think makes sense.

### New

- New test.

### Changed

- n/a

### Removed

- Obsolete/duplicate files.

## Testing and Reviewing

`npm run test` passes.


cdk/test/cdk.test.ts
```diff
@@ -1,17 +1,25 @@
...
+  const dataPlaneStack = new DataPlaneStack(app, util.stackName(util.StackId.DataPlane), {
+    vpc: networkStack.vpc,
+  });
+
+  // Assert for presence. See this list of resource types to find the strings to use here:
+  // https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html
+  const networkTemplate = Template.fromStack(networkStack);
+  networkTemplate.hasResourceProperties('AWS::EC2::VPC', {});
+  const dataPlaneTemplate = Template.fromStack(dataPlaneStack);
+  dataPlaneTemplate.hasResourceProperties('AWS::RDS::DBCluster', {}); // Aurora cluster.
```
justinebreuch, line 23: You can wait to add this but would be nice to check permissions on this

beekley: Good idea. I'll see if that's easy to do. If not, I'll add it in another PR.

kailamjeter (approved):

cdk/test/cdk.test.ts
beekley: I added a TODO for a future PR.

PR #13, opened 2022-05-19
Lambda  to write population data

justinebreuch:
## Description

Addresses: [shortcut story](https://app.shortcut.com/blueconduit/story/4867/serve-static-census-data-from-postgresql-db-in-aws)

Adds lambda function to insert rows into the demographics table in our main cluster

### New

lambda-layer-stack.ts
write-demographic-data-handler.ts

### Changed

Added lambda stack to data-plane-stack.ts


cdk/src/open-data-platform/lambda/lambda-layer-stack.ts
```diff
@@ -0,0 +1,63 @@
...
+import * as secretsmanager from 'aws-cdk-lib/aws-secretsmanager';
+import {Construct} from 'constructs';
+import * as path from 'path';
+
+interface SchemaProps {
+  cluster: rds.ServerlessCluster, vpc: ec2.IVpc, db: string,
+  credentialsSecret: secretsmanager.ISecret,
+}
+
+export class LambdaLayerStack extends Construct {
```
beekley, line 16: Could we name this something more descriptive about its purpose? Like `DataImportStack`.

cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
```diff
@@ -0,0 +1,233 @@
+// asset-input/src/open-data-platform/lambda/write-demographic-data-handler.js
```
beekley, line 1: When do you think this lambda will be invoked? If it's not during `cdk deploy` then it's not really part of the deployment and shouldn't be in the `cdk` directory.

cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
```diff
@@ -0,0 +1,233 @@
...
+const S3 = new AWS.S3();
+const rdsDataService = new AWS.RDSDataService();
+
+const GEO_ID = 'GEOID';
+const RACE_TOTAL = 'RaceTotal';
+const WHITE_POPULATION = 'Estimate!!Total:!!White alone';
+const BLACK_POPULATION = 'Estimate!!Total:!!Black or African American alone';
+
+/// Prepared SQL statements.
+const CREATE_DEMOGRAPHICS_TABLE =
```
beekley, line 15: Can you move this schema definition to [`schema.sql`](https://github.com/BlueConduit/open-data-platform/blob/main/cdk/src/open-data-platform/data-plane/schema.sql)? That should be the one spot where we define the schema for the DB.

cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
```diff
@@ -0,0 +1,233 @@
...
+const parseS3IntoDemographicsTableRow =
+  (s3Params: Object): Promise<Array<DemographicsTableRow>> => {
+    let results: DemographicsTableRow[] = [];
+    return new Promise(function (resolve, reject) {
+      let count = 0;
+      let fileStream = S3.getObject(s3Params).createReadStream();
+      fileStream.pipe(parse())
+                .on('data',
+                  (chunk) => {
+                    fileStream.pause();
```
beekley, line 37: Did this solve the problem we faced yesterday? If so, can you add a short comment explaining why we have this here?

cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
```diff
@@ -0,0 +1,233 @@
...
+const parse = require('csv-parser');
+const S3 = new AWS.S3();
+const rdsDataService = new AWS.RDSDataService();
+
+const GEO_ID = 'GEOID';
+const RACE_TOTAL = 'RaceTotal';
+const WHITE_POPULATION = 'Estimate!!Total:!!White alone';
+const BLACK_POPULATION = 'Estimate!!Total:!!Black or African American alone';
+
+/// Prepared SQL statements.
```
beekley, line 14: Does the triple `///` mean anything?

cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
```diff
@@ -0,0 +1,233 @@
...
+  '   black_percentage real,\n' +
+  '   white_percentage real, \n' +
+  '   PRIMARY KEY(census_geo_id) \n' +
+  ');';
+
+const DELETE_DEMOGRAPHICS_TABLE =
+  'DELETE FROM demographics WHERE census_geo_id IS NOT NULL';
+
+/// Reads the S3 CSV file and returns [DemographicsTableRow]s.
+const parseS3IntoDemographicsTableRow =
```
beekley, line 28: Is the purpose of this function to get *all* of the rows? If so, that's a lot of data to return in a single variable and defeats the purpose of processing the file as a stream.

But that's also not necessarily a problem as long as know that the files aren't too big to store in lambda memory. I'd call that out explicitly in the comment (and rename to include `Rows` plural).

cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
```diff
@@ -0,0 +1,233 @@
...
+   callback: APIGatewayProxyCallback): void => {
+    const secretArn = event['secretArn'];
+    const mainClusterArn = event['mainClusterArn'];
+
+    const s3Params = {
+      Bucket: 'opendataplatformapistaticdata',
+      Key: 'alabama_acs_data.csv'
+    };
+    // Read CSV file and write to demographics table.
+    parseS3IntoDemographicsTableRow(s3Params)
```
beekley, line 145: Can you use [`async/await`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) syntax here instead of `.then`? It makes code more simple and readable.

cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
beekley: Alternative suggestion: execute the `INSERT INTO` query for each chunk as its read. Instead of reading the entire file *then* executing the query on the whole set.

cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
```diff
@@ -0,0 +1,233 @@
...
+    let results: DemographicsTableRow[] = [];
+    return new Promise(function (resolve, reject) {
+      let count = 0;
+      let fileStream = S3.getObject(s3Params).createReadStream();
+      fileStream.pipe(parse())
+                .on('data',
+                  (chunk) => {
+                    fileStream.pause();
+                    // Only process 10 rows for now.
+                    if (count < 10) {
```
beekley, line 39: Is this limit intentionally left in?

cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
```diff
@@ -0,0 +1,233 @@
...
+                      let row =
+                        new DemographicsTableRowBuilder()
+                          .censusGeoId(chunk[GEO_ID])
+                          .totalPopulation(parseInt(chunk[RACE_TOTAL]))
+                          .blackPopulation(parseInt(chunk[BLACK_POPULATION]))
+                          .whitePopulation(parseInt(chunk[WHITE_POPULATION]))
+                          .build();
+                      count += 1;
+                      console.log(row);
```
beekley, line 48: This log is fine while there are only 10 rows, but will be messy after that. I'd suggest cutting it now.

cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
```diff
@@ -0,0 +1,233 @@
...
+                  })
+                .on('end', () => {
+                  console.log('Parsed ' + results.length + ' rows.');
+                  resolve(results);
+                });
+    });
+  }
+
+/// Parses rows and columns of SQL query into [SqlData].
+const parseSqlQuery = (data: Object): SqlData => {
```
beekley, line 65: Does anything call this function?

cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
```diff
@@ -0,0 +1,233 @@
...
+        callback('Error: RDS statement failed to execute');
+      } else {
+        console.log('Data is: ' + data);
+        callback(null, {statusCode: 200, body: JSON.stringify(data)});
+      }
+    });
+}
+/// Format [DemographicsTableRow] as SQL row.
+const formatPopulationTableRowAsSql = (row: DemographicsTableRow): string => {
+  let singleValue = [
```
beekley: nit: use `const` since `singleValue` doesn't change.

cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
```diff
@@ -0,0 +1,233 @@
...
+    const s3Params = {
+      Bucket: 'opendataplatformapistaticdata',
+      Key: 'alabama_acs_data.csv'
+    };
+    // Read CSV file and write to demographics table.
+    parseS3IntoDemographicsTableRow(s3Params)
+      .then(function (rows: Array<DemographicsTableRow>) {
+        let valuesForSql = rows.map(formatPopulationTableRowAsSql);
+
+        let insertRowsIntoDemographicsTableStatement =
```
beekley: Tributary uses the `@databases/pg` package for SQL operations. Their [`bulkInsert`](https://www.atdatabases.org/docs/pg-bulk) might be useful here so you don't have to do a bunch of string manipulation to craft a SQL query by hand.

beekley (commented):
If this review is too early, feel free to ignore it. I got the email that you opened the PR and dove in.

cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
justinebreuch: Done

cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
justinebreuch: Switched to two throughout

cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
justinebreuch: Yes but you can update to however many rows you want or I can add to the event variables?

cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
justinebreuch: Done!

cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
justinebreuch: Added a note about pausing for processing.

cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
justinebreuch: No moved it out

cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
justinebreuch: I did but ended up having to wrap it in a lot of nested promises which I think may have made it worse but at least it's all consistent now

kailamjeter (approved):

cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
justinebreuch: This is a good point. I'm going to try this and the db parsing in another PR!

cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
justinebreuch: Plan is to move to parallel directory called 'api' but in separate PR

cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
justinebreuch: Trying this in fast follow

cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
justinebreuch: Also for posterity: a chunk is just one row

PR #15, opened 2022-05-23
Migrate Vue FE to TypeScript.

kailamjeter:
## Description

Addresses: [migrate-fe-to-typescript](https://app.shortcut.com/blueconduit/story/4994/migrate-fe-to-typescript)

Move all FE files to TypeScript (from JavaScript). Uses lang="ts" declaration along with defineComponent to add type inference to components. Move .js files to .ts.

### New

- client/src/shims-vue.d.ts: defines how TypeScript should treat .vue files.

### Changed

- .js files to .ts
- defineComponent function call around vue components
- types to functions and variables

### Removed

N/A

## Testing and Reviewing

npm run serve


justinebreuch (approved):

client/src/components/MapLegend.vue
```diff
@@ -30,7 +32,10 @@ export default {
       required: true
     },
     bucketMap: {
-      type: Map,
+      // There is no constructor function for a Map of declared type, so use
+      // object here and cast to PropType of a Map<string, string>.
+      // See https://vuejs.org/guide/typescript/options-api.html#typing-component-props.
+      type: Map as PropType<Map<string, string>>,
```
beekley, line 38: Great comment.

client/src/components/MapLegend.vue
```diff
@@ -50,11 +55,11 @@ export default {
      * This is called when the component is mounted and any time the incoming
      * bucketMap is updated.
      */
-    createLegend() {
+    createLegend(): void {
```
beekley, line 58: Do you know if it's a best practice to put `void` for the return type here? We might have a lot of methods that don't return anything so we should be consistent (such as the below arrow function in a `forEach`).

beekley (approved):

PR #17, opened 2022-05-24
Use databases/pg to insert and delete rows instead of RDS service

justinebreuch:
## Description

Addresses: [https://app.shortcut.com/blueconduit/story/4867/serve-static-census-data-from-postgresql-db-in-aws](Shortcut)

Use databases/pg to write to db instead of AWS's RDS service. Has bulkInsert and helpful formatting helpers. 

## Testing and Reviewing

Using manual test events inside the lambda function. Will follow-up with Gateway call

### DIFF
--- a/cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
+++ b/cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
@@ -123,7 +123,6 @@ function parseS3IntoDemographicsTableRow(
             (dataRow) => {
               // Pause to allow for processing.
               fileStream.pause();
-              // Only process 10 rows for now.
               if (count < numberRowsToWrite) {
                 const row =
                     new DemographicsTableRowBuilder()
--- a/cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
+++ b/cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
@@ -1,158 +1,224 @@
 // asset-input/src/open-data-platform/lambda/write-demographic-data-handler.js
+import {SecretsManager} from '@aws-sdk/client-secrets-manager';
+import createConnectionPool, {ConnectionPool, ConnectionPoolConfig, Queryable, sql} from '@databases/pg';
+import {bulkInsert, BulkOperationOptions} from '@databases/pg-bulk';
 import {APIGatewayEvent} from 'aws-lambda';
 
 const AWS = require('aws-sdk');
 const parse = require('csv-parser');
 const S3 = new AWS.S3();
-const rdsDataService = new AWS.RDSDataService();
+const secretsmanager = new SecretsManager({})
 
 const GEO_ID = 'GEOID';
 const RACE_TOTAL = 'RaceTotal';
 const WHITE_POPULATION = 'Estimate!!Total:!!White alone';
 const BLACK_POPULATION = 'Estimate!!Total:!!Black or African American alone';
 
-// Prepared SQL statements.
-const DELETE_DEMOGRAPHICS_TABLE =
-    'DELETE FROM demographics WHERE census_geo_id IS NOT NULL';
+// Establishes connection with DB with the given ARN.
+function connectToDb(secretArn: string): Promise<ConnectionPool> {
+  return new Promise(async function(resolve, reject) {
+    try {
+      console.log('Fetching db credentials...');
+      const secretInfo =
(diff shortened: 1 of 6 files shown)


cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
```diff
@@ -1,158 +1,224 @@
...
+        tableName: `demographics`,
+        columnTypes: {
+          census_geo_id: sql`VARCHAR`,
+          total_population: sql`REAL`,
+          black_percentage: sql`REAL`,
+          white_percentage: sql`REAL`,
+        },
+      };
+
+  return Promise.all([bulkInsert({
```
justinebreuch, line 96: I don't quite understand why this makes me wrap it into Promise.all because bulkInsert returns a promise

cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
```diff
@@ -1,158 +1,223 @@
...
-    await rdsDataService.executeStatement(sqlParams,
-        function (err: Error, data: Object) {
-          if (err) {
-            console.log(err);
-            reject(err);
-          } else {
-            console.log('Data is: ' + JSON.stringify(data));
-            resolve(data);
-          }
+// Reads the S3 CSV file and returns [DemographicsTableRow]s.
```
kailamjeter, line 113: nit: let's use either // or /**/ for function comments (below you use the latter)

kailamjeter (approved):

cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
```diff
@@ -1,158 +1,223 @@
...
-                (error) => {
-                  reject(error);
-                })
-            .on('end', () => {
-              console.log('Parsed ' + results.length + ' rows.');
-              resolve(results);
-            });
+// Establishes connection with DB with the given ARN.
+function connectToDb(secretArn: string): Promise<ConnectionPool> {
+  return new Promise(async function(resolve, reject) {
```
beekley, line 19: I believe you can drop the use of a `new Promise` here by:
* Add the `async` keyword to this function declaration (`async function connectToDb`).
* Use `return pool` instead of `resolve(pool)`.
* Remove the try-catch blocks. `async will handle that for you, unless you want some specific error handling.

cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
```diff
@@ -1,158 +1,223 @@
...
+    try {
+      console.log('Fetching db credentials...');
+      const secretInfo =
+          await secretsmanager.getSecretValue({SecretId: secretArn});
+      const {host, port, username, password} =
+          JSON.parse(secretInfo.SecretString!);
+
+      const config: ConnectionPoolConfig = {
```
beekley, line 27: nit about style: you don't have to list both the key and variable names in an object if they are the same. E.g.
```
{ host, port, user: username, password }
```

cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
beekley: Also, I saw you comment on slack about reusing the code from schema.handler.ts. Good idea.

cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
beekley: Hmm yeah you should be able to just `return await bulkInsert`. Perhaps it's because the return type of `insertRows` is an array?

beekley (commented):
Some initial comments but I haven't read through it all yet.

cdk/src/open-data-platform/lambda/write-demographic-data-handler.ts
```diff
@@ -1,158 +1,223 @@
...
+      db = await connectToDb(secretArn);
 
+      // Remove existing rows before inserting new ones.
+      await deleteRows(db);
+      const data = await insertRows(db, rows);
       resolve({statusCode: 200, body: JSON.stringify(data)});
 
     } catch (error) {
-      console.log('Error:' + error);
+      console.log('Error:' + JSON.stringify(error));
```
beekley, line 180: `console.log('Error:', error);` might do what you want here.

beekley (approved):

PR #19, opened 2022-05-27
Kailajeter/sc 5004/implement nwm searchbar

kailamjeter:
## Description

Addresses: [A user can pick the visible data layer in the search bar](https://app.shortcut.com/blueconduit/story/5004/a-user-can-pick-the-visible-data-layer-in-the-search-bar)

Adds searchbar which shows data layer option and dropdown to select options. TBD on if dropdown will be needed but will leave until mocks are final.

Add state and datalayer classes which will be used to inject state data throughout the app. Plan is to modify existing components to use this as source of truth in a followup PR.

### New

- SearchBar, SearchBarOption, MapContentContainer components
- data_layer.ts, state.ts

### Changed

- App.vue

### Removed

- N/A

## Testing and Reviewing

npm run serve / npm run build.

## Screenshot

Screen Shot 2022-05-27 at 11 18 05 AM


client/src/model/state.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Class representing the current state of the app.
```
beekley, line 2: Could you elaborate on how this class is used. Does it represent a global state that gets passed down to (and updated by) components? Or is it local to a component?

beekley (approved):

client/src/model/state.ts
kailamjeter: Done!

client/src/components/SearchBar.vue
```diff
@@ -0,0 +1,102 @@
...
+    const state: State | undefined = inject(stateKey);
+
+    return {
+      state,
+    };
+  },
+  data() {
+    return {
+      options: [] as DataLayer[],
+      selectedOption: null as DataLayer | null,
```
justinebreuch, line 44: I'm not 100% sure but can you do "null as DataLayer?" ?

client/src/main.ts
```diff
@@ -1,8 +1,8 @@
...
-import router from './router'
+import { createApp } from 'vue';
+import App from './App.vue';
+import router from './router';
 
 /**
  * Starts up the application.
  */
-createApp(App).use(router).mount('#app')
+const app = createApp(App).use(router).mount('#app');
```
justinebreuch, line 8: How come this variable?

client/src/model/data_layer.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Class representing a data layer on the map.
```
justinebreuch, line 2: Optional nit: Let's not do "Class represents"... for comments since it seems repetitive. I would instead say what a data layer, conceptually, is

client/src/model/data_layer.ts
```diff
@@ -0,0 +1,14 @@
...
+ */
+export class DataLayer {
```
justinebreuch, line 4: Can this be an interface. It seems like in typescript you can just do interface DataLayer {name: string} and you can instantiate it? Pls correct me if I'm wrong bc I'm a total noob to typescript. I would try that or just name:string without the getter since it's not private right now anyway and doesn't seem necessary putting it behind getters?

client/src/model/state.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Class representing the global state of the app.
+ *
+ * Injected into components that need to read or write app data.
+ */
```
justinebreuch, line 5: Same comment here about class comments. What is a state going to hold beyond datalayers (if you know)?

justinebreuch (approved):

client/src/components/SearchBar.vue
kailamjeter: Required some other funky thing, but left a comment with explanation. Lmk if this doesn't make sense.

client/src/main.ts
kailamjeter: Was playing around and forgot to delete - nice catch!

client/src/model/data_layer.ts
kailamjeter: Done.

client/src/model/state.ts
kailamjeter: Done - left a TODO to elaborate when there are more fields.

client/src/model/data_layer.ts
kailamjeter: Great point - I think the downside is it won't have a constructor, it is more like defining a type for js objects, however for now you are right we don't need additional functionality. 

Depending on how we want to store data here it may make sense to convert it to class or implement the interface as a class in the future but for now this is fine.

PR #22, opened 2022-06-01
Kailajeter/sc 5004/a user can pick the visible data layer in

kailamjeter:
## Description

Addresses: [a-user-can-pick-the-visible-data-layer-in-the-search-bar](https://app.shortcut.com/blueconduit/story/5004/a-user-can-pick-the-visible-data-layer-in-the-search-bar)

Specifically:
- Add additional data layer to searchbar, implement searching + toggling layers

This PR adds inject / provide of state from App.vue -> MapView.vue + MapLegend.vue. The child components use state fields (currentDataLayer) to render information related to the displayed visual layer.

**Note** that right now this depends on a deep watcher on the state, which may be bad for performance in the long run. I've left a TODO to look into alternatives and found 1 which allowed watching on nested fields (see [this stackoverflow post](https://stackoverflow.com/questions/42133894/vue-js-how-to-properly-watch-for-nested-data)) but it was buggy used alongside mapbox's style functionality. 

### New

- lead_and_copper_violations_by_county_config.ts
- population_by_county_config.ts
- data_layer.ts

### Changed

- App.vue, MapView.vue, MapLegend.vue
- Small styling change in MapPopupContent.vue

### Removed

N/A

## Testing and Reviewing

npm run serve / npm run build

## Screencast

Trying a new extension here so let me know if this doesn't work and I'll try to find another or just add a screenshot:

https://watch.screencastify.com/v/RnlaNJIlmF1Ca5QMj7aE


client/src/App.vue
```diff
@@ -3,23 +3,56 @@
...
 import NavigationBar from './components/NavigationBar.vue';
-import { stateKey } from './injection_keys';
 import { DataLayer } from './model/data_layer';
 import { State } from './model/state';
+import axios from 'axios';
+import { populationByCountyDataLayer } from './data_layer_configs/population_by_county_config';
+import { leadAndCopperViolationsByCountyDataLayer } from './data_layer_configs/lead_and_copper_violations_by_county_config';
+import { stateKey } from './injection_keys';
+
+const OPEN_DATA_PLATFORM_API_URL = 'https://v2rz6wzmb7.execute-api.us-east-2.amazonaws.com/default';
```
beekley, line 16: Does this just point to your environment? Can you add a comment saying such?

client/src/components/MapLegend.vue
```diff
@@ -21,32 +23,27 @@ import { defineComponent, PropType } from 'vue';
  */
 export default defineComponent({
   name: 'MapLegend',
+  setup() {
+    const state: State | undefined = inject(stateKey);
```
beekley, line 27: Rather than allowing this to be `undefined`, is there a default state we could use?

beekley (approved):

PR #24, opened 2022-06-02
Make the tile server publicly accessible through CloudFront

beekley:
## Description

Addresses: [serve census block data](https://app.shortcut.com/blueconduit/story/4772/serve-census-block-data)

You can now make requests to the tile server through the internet.

### New

- New cloud function to do some URL manipulation.

### Changed

- The frontend stack now points CloudFront to the tile server.
- Relabeled the frontend as "client" in the docs, since the "frontend" in the CDK represents the AWS infrastructure that users directly interact with (e.g. CloudFront).

### Removed

- n/a

## Testing and Reviewing

I added a super basic presence test. Which I actually think is fairly useful. Running these tests shows that 1. the CDK config builds correctly and 2. it built stuff we expect to exist. Whether it all works together is TODO, but maybe not as valuable.

You can make a request to the tile server by adding `/tiles/v1` to the CloudFront domain name. For example, https://dugnnye6n16k1.cloudfront.net/tiles/v1/healthz points to my health check.


cdk/src/open-data-platform/app-plane/tileserver/tileserver.ts
```diff
@@ -35,14 +35,16 @@ export class TileServer extends Construct {
       cpu: 512, // measured in milliCPU; or 0.5 vCPU.
       taskImageOptions: {
         // Source: https://github.com/urbica/martin
-        image: ecs.ContainerImage.fromRegistry('urbica/martin'),
+        // TODO: Rename this reposity or tag to describe that it's the tile server, not the platform.
+        image: ecs.ContainerImage.fromRegistry('public.ecr.aws/m2w9u0k4/open-data-platform:latest'),
```
kailamjeter, line 39: is this just where the container is hosted on AWS?

cdk/src/open-data-platform/frontend/frontend-stack.ts
```diff
@@ -45,5 +51,35 @@ export class FrontendStack extends Stack {
       distribution: this.distribution,
       distributionPaths: ['/*'],
     });
+
+    // Add the tileserver to the Cloudfront distribution to make it publicly available.
+    // TODO: consider splitting this out into another file for organization.
```
kailamjeter, line 56: +1 to splitting the distribution and this out and adding this to props to pass to both separately in the future.

cdk/src/open-data-platform/app-plane/tileserver/tileserver.ts
beekley: Yeah, I had to manually upload it since it didn't exist in any of the public AWS repositories yet.

cdk/src/open-data-platform/frontend/frontend-stack.ts
beekley: ACK.

PR #25, opened 2022-06-03
Kailajeter/sc 5158/a user can use their device s location to

kailamjeter:
## Description

Addresses: [a-user-can-use-their-device-s-location-to-search-on-the-map](https://app.shortcut.com/blueconduit/story/5158/a-user-can-use-their-device-s-location-to-search-on-the-map)

Add address search (mapbox geocoder) to the searchbar. 

This does **not** include functionality for getting user's current location because that is not supported with the mapbox geocoder (at least in ts). I will add that functionality separately with the [GeolocateControl](https://docs.mapbox.com/mapbox-gl-js/api/markers/#geolocatecontrol).

### New

- Address searchbar

### Changed

- Searchbar component and state (add map to state).

### Removed

- N/A

## Testing and Reviewing

npm run serve + search by address

## Screencast

https://watch.screencastify.com/v/iP11OMkXFJLKpAow6rxk

Screen Shot 2022-06-03 at 12 01 33 PM


client/src/App.vue
```diff
@@ -25,7 +26,8 @@ export default defineComponent({
   setup() {
     // Create and provide default state. This is updated once API data is fetched.
     const initialDataLayer: DataLayer = null as unknown as DataLayer;
-    const state = reactive(new State([], initialDataLayer));
+    const initialMap: mapboxgl.Map = null as unknown as mapboxgl.Map;
```
beekley, line 29: What does the `as unknown` do here? It makes me feel like there's some weirdness going on with types. Maybe the `map` property of `State` should be nullable so this isn't necessary.

client/src/components/MapGeocoderWrapper.vue
```diff
@@ -0,0 +1,106 @@
...
+  setup() {
+    const state: State = inject(stateKey, State.default());
+
+    return {
+      state,
+    };
+  },
+  data() {
+    return {
+      geocoder: null as unknown as MapboxGeocoder,
```
beekley, line 39: Same as above regarding `as unknown` here.

client/src/components/MapGeocoderWrapper.vue
```diff
@@ -0,0 +1,106 @@
...
+    };
+  },
+  watch: {
+    'state.map': function(newMap: mapboxgl.Map) {
+      // Create geocoder when map is updated and geocoder is null. This is only triggered once per
+      // app load since the map is only created once.
+      if (newMap != null && this.geocoder == null) {
+        this.geocoder = new MapboxGeocoder({
+          accessToken: process.env.VUE_APP_MAP_BOX_API_TOKEN,
+          mapboxgl: mapboxgl as unknown as mapboxgl.Map,
```
beekley, line 50: Same as above regarding `as unknown` here.

client/src/components/MapGeocoderWrapper.vue
```diff
@@ -0,0 +1,106 @@
+<template>
+  <div v-show='visible' class='geocoder-content-expanded'>
+    <div class='geocoder' id='geocoder'></div>
+    <div class='search-button' @click='this.visible = !this.visible'>
+      <img src='@/assets/icons/search.svg' />
```
beekley, line 5: What's the source of this asset? Is it freely licensed?

beekley (changes requested):
LGTM except a concern about type stuff.

client/src/App.vue
kailamjeter: Yeah on a previous PR Justine left a comment about just putting null as Type instead of making the type like 'type | null' to avoid having to account for null values everywhere the field is accessed. I'll switch this to use Map | null so it is just nullable.

client/src/components/MapGeocoderWrapper.vue
kailamjeter: This one actually is necessary if we want to have this field be of type MapboxGeocoder without assigning an actual MapboxGeocoder value on init. This is because the inferred type will just be null if we don't do typecasting here (there is no way to explicitly declare type for data fields AFAIK). 

Alternatively we could initialize just an empty geocoder but its sort of misleading because it will not be linked to the map at all or show up on the page since we create the element when the actual geocoder is initialized - let me know if you prefer this approach.

client/src/components/MapGeocoderWrapper.vue
kailamjeter: This is necessary if we want a marker. This actually seems to be a bug with the geocoder code - it requires the type  mapboxgl.Map but expects just mapboxgl type (see example: https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-geocoder-outside-the-map/). 

I'll add a comment to clarify. Also, this is only required if we want the marker placed on the map when it zooms to a location. Otherwise, this can be left undefined.

client/src/components/MapGeocoderWrapper.vue
beekley: I'm a little concerned about using something that we don't know the license of, even temporarily. Can you message the author of the mocks to see if we can use it? Else I'm sure there's a freely licensed icon we can use.

client/src/App.vue
beekley: I believe making it nullable is the "right" way of doing things. Right now it's effectively nullable, but without any of the safeguards that making it explicitly nullable would provide.

This applies to each of the below cases too. It should either be nullable or have an empty, but valid instance of that type.

client/src/components/MapGeocoderWrapper.vue
beekley: What is the cost of having an empty geocoder? I think that sounds fine, as long as it's not some expensive operation.

client/src/components/MapGeocoderWrapper.vue
```diff
@@ -0,0 +1,113 @@
...
+      visible: false,
+    };
+  },
+  watch: {
+    'state.map': function(newMap: mapboxgl.Map) {
+      // Create geocoder when map is updated and geocoder is null. This is only triggered once per
+      // app load since the map is only created once.
+      if (newMap != null && this.geocoder == null) {
+        this.geocoder = new MapboxGeocoder({
+          accessToken: process.env.VUE_APP_MAP_BOX_API_TOKEN,
```
justinebreuch, line 53: Could you add a TODO here to use secrets manager?

client/src/components/MapGeocoderWrapper.vue
```diff
@@ -0,0 +1,113 @@
+<template>
+  <div v-show='visible' class='geocoder-content-expanded'>
```
beekley, line 2: I believe there should only be one root element for a component: https://eslint.vuejs.org/rules/no-multiple-template-root.html

(my IDE highlighted this for me, maybe because I have ESLint installed?)

justinebreuch (approved):

client/src/components/MapGeocoderWrapper.vue
beekley: I don't see how that example requires this casting. What is `mapboxgl` itself? It seems like a library, while the `MapboxGeocoder` says it wants "A mapbox-gl instance to use when creating Markers". Can we instantiate something to pass?

I'm also fine skipping markers for now unless you use them.

beekley (changes requested):
I'm still uncomfortable with the type check bypassing.

client/src/components/MapGeocoderWrapper.vue
kailamjeter: The example is just js, not typescript so the casting isn't necessary. I think there is a big with the mapbox geocoder types which require what you mention "an instance". I tried passing the map instance and got this issue:

https://stackoverflow.com/questions/57467842/getting-error-this-mapboxgl-marker-is-not-a-constructor-when-using-geocoder

I'll just remove it for now since it isn't on mocks and this is confusing.

client/src/App.vue
kailamjeter: SG will update others as well.

client/src/components/MapGeocoderWrapper.vue
kailamjeter: No cost - done!

client/src/components/MapGeocoderWrapper.vue
kailamjeter: Done - will look into getting linter working fully. I do have some hightlighting but that was not called out.

client/src/components/MapGeocoderWrapper.vue
kailamjeter: Reached out to author (Eric Schwartz) and he didn't know. Reaching out to Jackson and a few others and will resolve when I know for sure.

kailamjeter: Raanan confirmed that all icons in figma design system should be free to use.

beekley (approved):

PR #26, opened 2022-06-03
Prep for data import lambdas

beekley:
## Description

Addresses: [census block lsl data is stored in the db](https://app.shortcut.com/blueconduit/story/4764/census-block-data-is-stored-in-the-db)

This is some initial work to start importing data to the DB. Since we'll have multiple devs working on this at the same time, I wanted to check in the work we've done so far.

### New

- n/a

### Changed

- Added SRID so geo data can be displayed via the tile server.
- Moved the data import lambdas to the data layer stack, so it can share some code with another lambda that touches the DB.
- Some typescript-y cleanup.

### Removed

- n/a

## Testing and Reviewing

n/a

### DIFF
--- a/cdk/src/open-data-platform/data-plane/data-import/data-import-stack.ts
+++ b/cdk/src/open-data-platform/data-plane/data-import/data-import-stack.ts
@@ -33,7 +33,7 @@ export class DataImportStack extends Construct {
           CREDENTIALS_SECRET: credentialsSecret.secretArn,
           DATABASE_NAME: db,
         },
-        timeout: Duration.minutes(10),
+        timeout: Duration.minutes(5),
         bundling: {
           externalModules: ['aws-sdk'],
           nodeModules: ['csv-parser', '@databases/pg'],
@@ -70,7 +70,7 @@ export class DataImportStack extends Construct {
 
     // Import the data on CDK deploy.
     // TODO: enable this once it is ready to use.
-    if (true) {
+    if (false) {
       const init = new ResourceInitializer(this, 'ImportDemographicData', {
         initFunction: writeDemographicDataFunction,
       });
--- a/cdk/src/open-data-platform/data-plane/data-import/data-import-stack.ts
+++ b/cdk/src/open-data-platform/data-plane/data-import/data-import-stack.ts
@@ -6,7 +6,7 @@ import * as rds from 'aws-cdk-lib/aws-rds';
 import * as secretsmanager from 'aws-cdk-lib/aws-secretsmanager';
 import { Construct } from 'constructs';
 import * as path from 'path';
-import { ResourceInitializer } from '../../resource-initializer';
+import { ResourceInitializer } from '../../../resource-initializer';
 
 interface SchemaProps {
   cluster: rds.ServerlessCluster;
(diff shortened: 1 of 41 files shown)


cdk/src/open-data-platform/data-plane/data-import/data-import-stack.ts
```diff
@@ -40,32 +41,39 @@ export class DataImportStack extends Construct {
...
-    if (writeDemographicDataFunction.role?.roleArn != undefined) {
-      let role = iam.Role.fromRoleArn(
-        scope,
-        `{id} + -role`,
-        writeDemographicDataFunction.role.roleArn,
-      );
-      cluster.grantDataApiAccess(role);
+    // Import the data on CDK deploy.
+    // TODO: enable this once it is ready to use.
+    if (true) {
```
justinebreuch, line 73: Did you mean false here?

cdk/src/open-data-platform/data-plane/schema/schema.sql
```diff
@@ -28,12 +28,27 @@ CREATE TABLE IF NOT EXISTS demographics(
 );
 
 ALTER TABLE demographics
-    ADD COLUMN IF NOT EXISTS geom GEOMETRY(Polygon);
+    -- SRID 4326 maps the shape to latitude and longitude.
+    ADD COLUMN IF NOT EXISTS geom GEOMETRY(MultiPolygon, 4326);
```
justinebreuch, line 32: Can you move to line 26?

cdk/src/open-data-platform/data-plane/schema/schema.sql
```diff
@@ -28,12 +28,27 @@ CREATE TABLE IF NOT EXISTS demographics(
...
 
+CREATE TABLE IF NOT EXISTS lead_service_lines(
+    pws_id varchar(255) NOT NULL,
+    lead_connections_count real,
+    PRIMARY KEY(pws_id)
+    );
+
+ALTER TABLE lead_service_lines
+    -- SRID 4326 maps the shape to latitude and longitude.
+    ADD COLUMN IF NOT EXISTS geom GEOMETRY(MultiPolygon, 4326);
```
justinebreuch, line 46: Perhaps make this number a constant explaining SRID 4326 maps the shape to latitude and longitude?

justinebreuch (approved):

cdk/src/open-data-platform/data-plane/data-import/data-import-stack.ts
```diff
@@ -40,32 +41,39 @@ export class DataImportStack extends Construct {
       },
```
justinebreuch, line 41: Could you change the timeout back to 5? I'm realizing the timeouts I saw were inside a library in the handler not on the handler itself - thanks!

cdk/src/open-data-platform/data-plane/data-import/data-import-stack.ts
beekley: Good catch, yes.

cdk/src/open-data-platform/data-plane/schema/schema.sql
beekley: No, because someone might have already created the `demographics` table and therefore this column change won't execute. So we need to add it in a separate query.

cdk/src/open-data-platform/data-plane/data-import/data-import-stack.ts
beekley: Yup, done.

cdk/src/open-data-platform/data-plane/schema/schema.sql
beekley: Good idea. Done.

PR #28, opened 2022-06-03
Lambda to write to the water systems table 

justinebreuch:
## Description

Addresses: [Shortcut story](https://app.shortcut.com/blueconduit/story/4768/water-system-lsl-data-is-stored-in-the-db)

Writes geometries to water_systems and moves some database helpers out of other lambdas

## Testing and Reviewing

Test events in the AWS console


cdk/src/open-data-platform/data-plane/data-import/data-import-stack.ts
```diff
@@ -71,7 +91,7 @@ export class DataImportStack extends Construct {
     // Import the data on CDK deploy.
     // TODO: enable this once it is ready to use.
     if (false) {
```
beekley, line 93: How do we invoke the data import lambda(s)? Manually?

cdk/src/open-data-platform/data-plane/schema/schema.handler.ts
```diff
@@ -108,6 +100,23 @@ async function runSchemaFile(file: string, db: ConnectionPool) {
...
+  const { host, port, username, password } = await getCredentials(CREDENTIALS_SECRET);
+  return {
+    host: host,
+    port: port,
+    user: username,
+    password: password,
+  };
+}
+
+export async function connectToDb(config?: ConnectionPoolConfig): Promise<ConnectionPool> {
```
beekley, line 114: Why does this function exist when we have `database` already? Is there a case when `config` can be null?

beekley (approved):

cdk/src/open-data-platform/data-plane/schema/schema.handler.ts
justinebreuch: Since we don't want the handlers to have to do this credential fetching themselves (it's the same every time), it's null. Happy to update this with any suggestions you may have

beekley: When does the credential get fetched if not during the handler execution?

justinebreuch: Through createDatabaseConfig(). Should I not even allow the option to pass a config?

beekley: Does that get called whenever `handler` runs? Could you walk me through how this code means we don't have to fetch the credentials every time?

justinebreuch: I added some comments to explain what's happening. The data writing handlers (demographics and water systems) call connectToDb() so that they don't repeat the same getCredential and database() code. The schema handler's handle() function also needs access to the db config so that is its own helper as well

PR #29, opened 2022-06-06
Create parcel-level data table.

beekley:
## Description

Addresses: [Parcel level data in DB](https://app.shortcut.com/blueconduit/story/5146/parcel-level-lsl-prediction-data-for-one-city-is-stored-in-the-db)

Just the table. The lambda that imports data here will come once we have that data. I'm trying to keep PRs short.

### New

- `parcels` table.

### Changed

- Removed the constants, since they weren't working anyway. SQL doesn't let you define constants like that unless it's in a transaction or something.
- Section header comments looks different. I'm not set on this, but wanted to make them stand out more.

### Removed

- n/a

## Testing and Reviewing

n/a

### DIFF
--- a/cdk/src/open-data-platform/data-plane/schema/schema.sql
+++ b/cdk/src/open-data-platform/data-plane/schema/schema.sql
@@ -56,14 +56,13 @@ CREATE INDEX IF NOT EXISTS geom_index
 -- Parcel-level data
 
 CREATE TABLE IF NOT EXISTS parcels (
-    id serial PRIMARY KEY
-);
-ALTER TABLE parcels
+    id serial PRIMARY KEY,
     -- TODO: consider standardizing addresses into a PostGIS type once we have data.
     -- https://postgis.net/docs/manual-2.5/Address_Standardizer.html
-    ADD COLUMN IF NOT EXISTS address text,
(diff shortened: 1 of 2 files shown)


cdk/src/open-data-platform/data-plane/schema/schema.sql
```diff
@@ -34,27 +32,45 @@ CREATE TABLE IF NOT EXISTS demographics(
...
--- Roles and Grants
+-- Parcel-level data
+
+CREATE TABLE IF NOT EXISTS parcels (
+    id serial PRIMARY KEY
+);
+ALTER TABLE parcels
+    -- TODO: consider standardizing addresses into a PostGIS type once we have data.
+    -- https://postgis.net/docs/manual-2.5/Address_Standardizer.html
+    ADD COLUMN IF NOT EXISTS address text,
```
justinebreuch, line 64: How come these statements can't be in the CREATE TABLE section?

justinebreuch (approved):

cdk/src/open-data-platform/data-plane/schema/schema.sql
beekley: No reason. I've merged them.

PR #30, opened 2022-06-06
Add geolocate control to map.

kailamjeter:
## Description

Addresses: [a-user-can-use-their-device-s-location-to-search-on-the-map](https://app.shortcut.com/blueconduit/story/5158/a-user-can-use-their-device-s-location-to-search-on-the-map)

Adds Geolocate control to map which allows user to zoom to their current location.

### New

- Geolocate control on map

### Changed

- MapView

### Removed

N/A

## Testing and Reviewing

npm run serve

## screencast

https://watch.screencastify.com/v/bQsadTt5gpDQHtJzMlab

### DIFF
--- a/client/src/components/MapView.vue
+++ b/client/src/components/MapView.vue
@@ -130,6 +130,29 @@ export default defineComponent({
     },
 
     /**
+     * Set up map controls.
+     *
+     * This currently consists of the GeolocateControl which sets up
+     * functionality to pan to the user's current location.
+     */
+    setUpControls(): void {
+      if (this.map == null) return;
+
+      // Add geolocate control to the map.a
+      this.map.addControl(
+        new mapboxgl.GeolocateControl({
+          positionOptions: {
+            // If enabled, gets the best possible results. Can result in slower response times or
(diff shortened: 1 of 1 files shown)


client/src/components/MapView.vue
```diff
@@ -123,6 +123,29 @@ export default defineComponent({
...
+     * Set up map controls.
+     *
+     * This currently consists of the GeolocateControl which sets up
+     * functionality to pan to the user's current location.
+     */
+    setUpControls(): void {
+      if (this.map == null) return;
+
+      // Add geolocate control to the map.a
+      this.map.addControl(
```
justinebreuch, line 136: Don't feel like you need to address in this PR, but does this work on mobile?

kailamjeter: Not exactly sure from documentation, it just says that 'not all browsers support geolocation'. For mobile, it will still be opened in a web browser so I think it should be the same as whether that browser is supported in general? Not exactly sure though.

justinebreuch (approved):

PR #31, opened 2022-06-07
Add DataImportApiStack to import data sources into db via API

justinebreuch:
## Description

Addresses: [Initialize API](https://app.shortcut.com/blueconduit/story/3679/initialize-api)

Adds Stack to provision REST API for data imports. Adds lambda integrations such that endpoints can invoke lambda functions.

Considering changing process.env variables in lambda to query parameters (in separate PR)

### New

DataImportApiStack

## Testing and Reviewing

AWS console test events


beekley:
My first thought is that this isn't necessary. If us devs are the users of this endpoint, I'd rather invoke the lambda via the AWS management console than build our own interface.

beekley:
I'm closing this as obsolete. Feel free to reopen as you think necessary.

PR #34, opened 2022-06-08
Cdk error

justinebreuch:
## Description

Addresses: [Display Water System data on the map](https://app.shortcut.com/blueconduit/story/4784/display-water-system-data-on-the-map)

Adds the water_system tile vectors as a data layer to the map. 

### New

water_systems_config.ts 

Also fixed an error in the CDK code that kept me from deploying. Sorry it's part of this PR. 

### Changed

Some new interfaces in the data_layer to add sources for both geoJson and vectors. 

## Testing and Reviewing

The black is a result of non-number inputs. Going to figure that out separately.

Screen Shot 2022-06-08 at 4 04 30 PM

### DIFF
--- a/client/src/components/MapView.vue
+++ b/client/src/components/MapView.vue
@@ -130,14 +130,18 @@ export default defineComponent({
       if (this.map == null) return;
 
       this.state.map = this.map;
-      this.state.dataLayers?.forEach(layer => {
-        const source: GeoJSONSourceRaw = {
-          type: 'geojson',
-          data: layer.data,
-        };
-        this.map?.addSource(layer.id, source);
-        this.map?.addLayer(layer.styleLayer);
-      });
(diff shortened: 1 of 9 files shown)


client/src/components/MapView.vue
```diff
@@ -70,7 +72,7 @@ export default defineComponent({
      */
     updateMapOnStateChange(newState: State): void {
       if (this.map == null) {
-        if (newState.currentDataLayer?.data != null) {
+        if (newState.currentDataLayer?.source != null) {
```
justinebreuch, line 75: Kaila, what does this do?

kailamjeter (approved):

client/src/components/MapView.vue
justinebreuch: Updated to toggle on type

PR #37, opened 2022-06-09
Add navigation control (zoom in / out) to map.

kailamjeter:
## Description

Addresses: [a-user-can-search-for-an-address-in-a-searchbar](https://app.shortcut.com/blueconduit/story/5176/a-user-can-search-for-an-address-in-a-searchbar)

Add navigation controls (zoom in / out).

### New

- Navigation control on map.

### Changed

N/A

### Removed

N/A

## Testing and Reviewing

npm run serve

## screenshot
Screen Shot 2022-06-09 at 12 58 50 PM

### DIFF
--- a/client/src/components/MapView.vue
+++ b/client/src/components/MapView.vue
@@ -4,14 +4,13 @@
 </template>
 
 <script lang='ts'>
-import mapboxgl, { AnySourceData } from 'mapbox-gl';
-import mapbox, { GeoJSONSourceRaw, LngLatLike, MapLayerMouseEvent } from 'mapbox-gl';
+import mapboxgl from 'mapbox-gl';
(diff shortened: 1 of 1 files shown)


client/src/components/MapView.vue
```diff
@@ -150,6 +149,9 @@ export default defineComponent({
           trackUserLocation: true,
         }),
       );
+
+      // Add zoom in / zoom out buttons to map.
+      this.map.addControl(new mapboxgl.NavigationControl());
```
justinebreuch, line 154: Do you need to remove the other code you added?

kailamjeter: Do you mean the other zoom handler? never merged it, just closed out the PR so no need to remove.

PR #43, opened 2022-06-17
Fix water systems and epa violations handlers to write all rows

justinebreuch:
Fix water systems and epa violations handlers to write all rows

## Description

Addresses: [Display EPA violations data on the map](https://app.shortcut.com/blueconduit/story/5481/display-epa-violations-data-on-the-map)

Updates both water systems and violations data handlers to use a new better-supported library for writing to the db (pg not @databases/pg).

Allows passing in start index for writing rows to make it easier to run this lambda multiple times without too much editing.

With both handlers fixed, I could create a violations_count view that aggregated violations at the water system level. 

### New

New view inside schema.sql

### Changed

Both handlers have been updated and refactored. 

### Removed

Use of @database/pg

## Testing and Reviewing

AWS test events

PR #44, opened 2022-06-23
Uncommited changes from Pull 43

justinebreuch:
This is a commit that didn't make it into https://github.com/BlueConduit/open-data-platform/pull/43. I'm unsure what happened it,  but it includes the responses to the last wave of feedback.

### DIFF
--- a/cdk/src/open-data-platform/data-plane/data-import/write-water-systems-data-handler.handler.ts
+++ b/cdk/src/open-data-platform/data-plane/data-import/write-water-systems-data-handler.handler.ts
@@ -17,7 +17,7 @@ const Batch = require('stream-json/utils/Batch');
 
 // Number of rows to write at once.
 const BATCH_SIZE = 10;
-const DEFAULT_NUMBER_ROWS_TO_INSERT = 30000;
+const DEFAULT_NUMBER_ROWS_TO_INSERT = 10000;
 const NUMBER_ROWS_BUMPER = 300;
 const SCHEMA = 'public';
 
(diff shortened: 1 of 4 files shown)


cdk/src/open-data-platform/data-plane/data-import/write-water-systems-data-handler.handler.ts
```diff
@@ -194,11 +215,20 @@ export async function handler(_: APIGatewayProxyEvent): Promise<APIGatewayProxyR
     // See https://docs.aws.amazon.com/rds/index.html.
     const rdsService = new AWS.RDSDataService();
 
+    try {
+      const rowsInDb = (await determineRowsWritten(rdsService)).records[0][0].longValue;
+      startIndex = Math.max(parseInt(rowsInDb) - NUMBER_ROWS_BUMPER, startIndex);
```
beekley, line 220: Same comment as before. I'm 99% sure you can't assume that N rows in the DB are the first N rows in the file. Is that acceptable?

beekley (changes requested):

cdk/src/open-data-platform/data-plane/data-import/write-water-systems-data-handler.handler.ts
justinebreuch: If you only pass in numberOfRows that you know it can handle and it does not time out, it will do the first numberOfRows to write in the file, right? It doesn't matter if 800-900 get committed before 900-1000 as long as 10k rows go in

beekley: SGTM

justinebreuch: Cool resolving and submitting

PR #45, opened 2022-06-24
Switch water systems importer to use factory.

beekley:
## Description

The factory abstracts a lot of the code from the current water systems handler out into something generic. This PR takes the improvements from other recent PRs and shows how it can be used to import rows into the DB.

### Changed

- water system importer uses the factory now.
- Batches with errors are split into individual rows and re-imported.
- Logs are grouped by batch, e.g. you can search `batch414` and see everything that happened in that batch.

## Testing and Reviewing

I deployed the tested the handler. It successfully imported all 26010 rows in 10 minutes in one lambda invocation.


cdk/src/open-data-platform/data-plane/data-import/handler-factory.ts
```diff
@@ -48,59 +61,79 @@ export const geoJsonHandlerFactory = (
...
+            skippedRowCount += batchSize;
+            console.log(
+              `Skipping batch of ${rows.length} rows. ${skippedRowCount}/${rowOffset} offset rows have been skipped so far.`,
+            );
+            return;
+          }
+
+          console.log(`Processing batch of ${rows.length} rows.`);
+          const promise = callback(rows, db);
+          promises.push(promise);
```
justinebreuch, line 76: How come this is pushing to the list and then awaiting it? I think awaiting this would take away the parallelism and really slow down the writes / reads

cdk/src/open-data-platform/data-plane/data-import/handler-factory.ts
```diff
@@ -32,13 +39,19 @@ export const geoJsonHandlerFactory = (
...
     new Promise((resolve, reject) => {
       const batchSize = 10;
-      let numberRows = 0;
+      let processedRowCount = 0;
+      let skippedRowCount = 0;
 
       console.log('Starting to process file in S3:', s3Params);
 
+      // Store all of the promises returned by the callback so the Lambda waits for their resolution
+      // before exiting. They will start to execute before they are awaited and run asynchonously
```
justinebreuch, line 51: If you await the promise it's not running parallel, no?

cdk/src/open-data-platform/data-plane/data-import/handler-factory.ts
beekley: I added a comment describing why.

cdk/src/open-data-platform/data-plane/data-import/handler-factory.ts
```diff
@@ -1,44 +1,78 @@
...
    *
-   * TODO: track how many rows were processed and avoid re-importing the old rows.
+   * TODO: track how many batches are running in parallel and stop after a limit.
    *
    * @param db - Connection pool to the DB, made available to the callback.
    * @returns
    */
-  const readGeoJsonFile = (db: ConnectionPool): Promise<number> =>
+  const readGeoJsonFile = (
+    db: AWS.RDSDataService,
```
justinebreuch, line 59: Could you add @param docs for these for consistency?

cdk/src/open-data-platform/data-plane/data-import/handler-factory.ts
```diff
@@ -48,59 +82,112 @@ export const geoJsonHandlerFactory = (
...
   /**
    * Constructed handler that imports GeoJSON data to a DB.
    */
-  const handler = async (_: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
-    const db = await connectToDb();
-    if (db == undefined) {
-      throw Error('Unable to connect to db');
-    }
-    let numberRows = 0;
+  const handler = async (event: ImportRequest): Promise<APIGatewayProxyResult> => {
```
justinebreuch, line 167: Docs here too

cdk/src/open-data-platform/data-plane/data-import/handler-factory.ts
```diff
@@ -1,44 +1,78 @@
...
    * processing of the data itself, but provides a DB connection to the callback to process it.
    *
-   * TODO: track how many rows were processed and avoid re-importing the old rows.
+   * TODO: track how many batches are running in parallel and stop after a limit.
    *
    * @param db - Connection pool to the DB, made available to the callback.
    * @returns
    */
-  const readGeoJsonFile = (db: ConnectionPool): Promise<number> =>
+  const readGeoJsonFile = (
```
kailamjeter, line 58: a bit confused about the structure of this - what does wrapping this in the const geoJsonHandlerFactory do and why put this in a variable instead of function signature?

cdk/src/open-data-platform/data-plane/data-import/handler-factory.ts
```diff
@@ -48,59 +82,112 @@ export const geoJsonHandlerFactory = (
...
+            );
+            return;
+          }
+
+          // Start processing the row.
+          console.log(`batch${id}: Processing batch of ${rows.length} rows.`);
+          const promise = callback(rows, db);
+          promises.push(promise);
+
+          // Reprocess errored rows individually. This may catch transient errors, or at least let
```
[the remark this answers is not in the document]
kailamjeter: +1

cdk/src/open-data-platform/data-plane/data-import/handler-factory.ts
```diff
@@ -48,59 +82,112 @@ export const geoJsonHandlerFactory = (
...
+            // Handle in the promise.
+          }
+
+          // Stop reading stream if this would exceed number of rows to write.
+          // This check is done at the end because it doesn't know how many rows are currently
+          // being processed in parallel.
+          if (processedRowCount > rowLimit) {
+            console.log(`batch${id}: Stopping after passing row limit:`, rowLimit);
+            pipeline.destroy();
+            return;
```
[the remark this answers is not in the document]
kailamjeter: +1 I think this will go right to ending the stream if pipeline.destroy is called

cdk/src/open-data-platform/data-plane/data-import/handler-factory.ts
```diff
@@ -48,59 +82,112 @@ export const geoJsonHandlerFactory = (
...
+              rePromise.catch((reason) => console.log(`batch${id}: Reprocessing failed`, reason));
+              promises.push(rePromise);
             }),
-          );
-          pipeline.resume();
+          ); // Supress unhandled rejections here.
+
+          try {
+            // Await this batches so we know whether to increment the row count.
+            // This does not block other batches from being processed in parallel.
```
kailamjeter, line 118: is this just so that the logging prints accurate # of rows processed? IIUC this might be causing the processing of more rows than specified since we await to increment processedRowCount while other batches are initiated.

cdk/src/open-data-platform/data-plane/data-import/write-water-systems-data-handler.handler.ts
```diff
@@ -324,7 +90,7 @@ class WaterSystemsTableRowBuilder {
     return [
       {
         name: 'pws_id',
-        value: { stringValue: this._row.pws_id },
+        value: { stringValue: this._row.pws_id }, // FIGURE OUT WHY THIS IS EMPTY
```
kailamjeter, line 93: is this comment outdated?

cdk/src/open-data-platform/data-plane/data-import/handler-factory.ts
kailamjeter, line 167: we can cover this in standup so feel free to resolve after, but wondering how this gets invoked?

cdk/src/open-data-platform/data-plane/data-import/handler-factory.ts
beekley: I guess only the `handler` needs to be wrapped in the `geoJsonHandlerFactory`, while the other functions could receive `s3Params` and `callback` through their arguments. Let me see if that makes this more readable and addresses your question.

As for why declare it as variable instead of function: just a style preference. But I'm happy to switch to functions instead.

cdk/src/open-data-platform/data-plane/data-import/handler-factory.ts
beekley: Good point. I'll try to address that. My main concern was that I don't want to count rows before they're inserted in case they ultimately fail. But that's minor.

cdk/src/open-data-platform/data-plane/data-import/handler-factory.ts
beekley: `geoJsonHandlerFactory` returns a `handler` function, which `write-water-systems-data-handler.ts` uses as its handler, which is invoked like any other lambda handler.

I'll talk through it at standup.

cdk/src/open-data-platform/data-plane/data-import/handler-factory.ts
```diff
@@ -1,44 +1,78 @@
...
+ * exceed it due to how it proccesses rows in parallel. Example:
+ *
+ * {
+ *   "rowOffset": 0,
+ *   "rowLimit": 27000,
+ *   "batchSize": 10
+ * }
+ *
  * @param s3Params - Details for how to get the GeoJSON file.
  * @param callback - Function that operates on a single element from the file.
```
kailamjeter, line 42: nit: update wording to convey that the callback operates on a whole batch instead of a single row

cdk/src/open-data-platform/data-plane/data-import/handler-factory.ts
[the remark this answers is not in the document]
beekley: I tried that but I think it made readability worse. The function signature alone is more lines of code than what goes into the `promise.catch` here:

```
const reprocessRow = async (
          db: AWS.RDSDataService,
          row: any[],
          batchId: string,
          callback: ImportCallback,
          promises: Promise<any>[],
        ): Promise<void> => {
          rows.forEach((row) => {
            console.log(`${id}: Reprocessing errored row individually.`, row.value?.properties);
            const rePromise = callback([row], db);
            rePromise.catch((reason) => console.log(`${id}: Reprocessing failed`, reason));
            promises.push(rePromise);
          });
        };
```

cdk/src/open-data-platform/data-plane/data-import/handler-factory.ts
```diff
@@ -48,59 +82,112 @@ export const geoJsonHandlerFactory = (
...
-                console.log('Stopping after row write limit:', rowLimit);
-                pipeline.destroy();
-                return;
-              }
-              // Multiple rows may be processed concurrently, so increment and check the limit
-              // before starting the processing step.
-              ++numberRows;
-              await callback(row, db);
+        .on('data', async (rows: any[]) => {
+          const id = bactchId++;
```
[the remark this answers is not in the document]
beekley: Done.

cdk/src/open-data-platform/data-plane/data-import/handler-factory.ts
beekley: Done.

cdk/src/open-data-platform/data-plane/data-import/handler-factory.ts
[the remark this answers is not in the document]
beekley: Sort of obsolete. I moved this if block up so the return statement is necessary to prevent the code below it from executing.

cdk/src/open-data-platform/data-plane/data-import/handler-factory.ts
beekley: I added a variable `inProcessRowCount`, which stores the number of rows being processed. This means we can do a check before processing a new batch rather than after.

cdk/src/open-data-platform/data-plane/data-import/handler-factory.ts
beekley: Done.

cdk/src/open-data-platform/data-plane/data-import/handler-factory.ts
```diff
@@ -1,108 +1,199 @@
...
+// These are defined here because the geoJsonHandlerFactory function signature was unreadable.
+type ImportCallback = (rows: any[], db: AWS.RDSDataService) => Promise<void>;
+type ImportHandler = (event: ImportRequest) => Promise<APIGatewayProxyResult>;
+
 const S3 = new AWS.S3();
 
 /**
  * Builds a lambda handler that parses a GeoJSON file and executes a callback.
+ *
+ * The callback takes in a set of rows and should insert them into the DB using the provided db
```
kailamjeter, line 32: nit: even though this is currently only used for insert, nothing about this is specific to inserting right? if not maybe we should keep documentation general - i.e. "The callback takes in a set of rows and a connection to the db and can do anything with those rows"... maybe not exactly that wording but you get the gist!

cdk/src/open-data-platform/data-plane/data-import/handler-factory.ts
```diff
@@ -1,108 +1,199 @@
...
-import { connectToDb } from '../schema/schema.handler';
+import { AnyLengthString } from 'aws-sdk/clients/comprehendmedical';
 
 // These libraries don't have types, so they are imported in a different way.
 const { chain } = require('stream-chain');
 const { streamArray } = require('stream-json/streamers/StreamArray');
 const Batch = require('stream-json/utils/Batch');
 const Pick = require('stream-json/filters/Pick');
 
+export interface ImportRequest {
```
kailamjeter, line 11: same comment here as below - since the handler is not insert-specific, could we just call this geojsonparsingrequest?

kailamjeter (approved):

cdk/src/open-data-platform/data-plane/data-import/handler-factory.ts
beekley: Good point. I've removed the "insert" or "import" language and replaced it with "process".

cdk/src/open-data-platform/data-plane/data-import/handler-factory.ts
beekley: I think that's a fair point, but I kept it as-is here since the only use for the callback right now is inserts. So it tells the user what they should put in the callback. But I'm up to change it if we ever use this for another purpose.

justinebreuch (approved):

PR #49, opened 2022-06-28
Read tileserver instead of GeoJson file for violations data layer + add query params for data layer 

justinebreuch:
## Description

Addresses: [Display violations data on map](https://app.shortcut.com/blueconduit/story/5649/display-violations-data-on-map)

Uses tileserver instead of geoJSON file for violations layer. 

Adds router logic so that layers can be toggled through the query params. 

Updates some small water systems data based on recent changes (relevant for violations because violations is water-systems based for now and is violations per service line)

### New

Routing logic for query params

### Changed

Data layer configs to use the tileserver instead of geojson files 

### Removed

Population layer for now to simplify data sources

## Testing and Reviewing

Local server


client/src/components/MapView.vue
```diff
@@ -39,6 +40,11 @@ export default defineComponent({
       popup: null as mapboxgl.Popup | null,
     };
   },
+  computed: {
+    routerLayer () {
```
kailamjeter, line 44: can you leave a comment as to what routerLayer represents

client/src/components/MapView.vue
```diff
@@ -95,9 +106,7 @@ export default defineComponent({
     updateMapOnDataLayerChange(newDataLayer: DataLayer | null, oldDataLayer: DataLayer | null): void {
       if (this.map == null) {
         const source = newDataLayer?.source;
-        const dataLoaded =
-          (source?.type == DataSourceType.GeoJson && source?.data != null) ||
-          (source?.type == DataSourceType.Vector && source?.tiles != null);
+        const dataLoaded = source?.type == DataSourceType.Vector && source?.tiles != null;
```
kailamjeter, line 109: nit: I think we can remove the source.type check since all data sources will now be vector

client/src/components/MapView.vue
```diff
@@ -223,11 +232,20 @@ export default defineComponent({
...
+    this.updateMapOnDataLayerChange(this.state.currentDataLayer, null);
+  },
   watch: {
+    // Listens to app state to toggle layers.
     'state.currentDataLayer': function(newDataLayer: DataLayer, oldDataLayer: DataLayer) {
       this.updateMapOnDataLayerChange(newDataLayer, oldDataLayer);
     },
-  },
+    // Listens to query param to toggle layers.
+    'routerLayer': function(newLayer: string) {
```
kailamjeter, line 245: does this not get triggered when you update URL in setDataLayerVisibility (/wouldn't that cause a loop)?

client/src/data_layer_configs/lead_service_lines_by_water_systems_config.ts
```diff
@@ -32,12 +32,12 @@ const LEGEND_COLOR_MAPPING = [
 const leadConnectionLegendInterpolation = [
   'interpolate',
   ['linear'],
-  ['get', 'lead_connections_count'],
+  ['/', ['get', 'lead_connections_count'], ['get', 'service_connections_count']],
```
kailamjeter, line 35: what does this do?

justinebreuch: Added comment: provides ratio of lead lines : all lines

client/src/components/MapView.vue
justinebreuch: Removed in the condition. We have to keep it overall because the types for mapbox require type

client/src/components/MapView.vue
justinebreuch: Done!

client/src/components/MapView.vue
justinebreuch: Basically the only thing is  router.currentRoute.value.query.layer != styleLayerId in setDataLayerVisibility(). We need both the app state and the router url to update the data layer and both need to stay in style. So this one setter needs to update both.

kailamjeter: Gotcha, thanks!

kailamjeter (approved):

client/src/App.vue
```diff
@@ -3,20 +3,20 @@
...
-import { leadAndCopperViolationsByCountyDataLayer } from './data_layer_configs/lead_and_copper_violations_by_county_config';
-import { stateKey } from './injection_keys';
-import { leadServiceLinesByWaterSystemLayer } from './data_layer_configs/water_systems_config';
+
+import { leadAndCopperViolationsByCountyDataLayer } from './data_layer_configs/lead_and_copper_violations_by_water_system_config';
+import { leadServiceLinesByWaterSystemLayer } from './data_layer_configs/lead_service_lines_by_water_systems_config';
 import { populationDataByCensusBlockLayer } from './data_layer_configs/population_by_census_block_config';
+import { stateKey } from './injection_keys';
+
+const DATA_LAYERS = [leadServiceLinesByWaterSystemLayer, leadAndCopperViolationsByCountyDataLayer, populationDataByCensusBlockLayer];
```
beekley, line 16: could you make this a map or enum, so we don't have to refer to it by index?

client/src/components/MapPopupContent.vue
```diff
@@ -64,9 +64,24 @@ export default defineComponent({
       this.displayedProperties.clear();
 
       for (let entry of this.featureProperties) {
+        console.log(this.featureProperties);
```
beekley, line 67: If this log is intentional, can you add some string before it to put it into context?

client/src/components/MapView.vue
```diff
@@ -59,6 +69,11 @@ export default defineComponent({
       if (this.map == null) return;
 
       this.map?.setLayoutProperty(styleLayerId, 'visibility', visible ? 'visible' : 'none');
+
+      // Update the router params when toggling layers to visible.
+      if (visible && router.currentRoute.value.query.layer != styleLayerId) {
+        router.push({ query: Object.assign({}, router.currentRoute.value.query, { layer: styleLayerId }) });
```
beekley, line 75: This is great.

client/src/components/MapPopupContent.vue
justinebreuch: Whoops good catch thanks!

client/src/components/MapView.vue
justinebreuch: Thanks!

client/src/components/MapView.vue
justinebreuch: Just kidding once I removed it complained that only Vector types have tiles so the condition had to be there

PR #53, opened 2022-06-29
Use lambda handler factory for data imports

beekley:
## Description

Removes redundant code via the factory.

### Changed

- Switches demographic and parcel data to use the handler factory.
- Moves model data to its own dir.

## Testing and Reviewing

I haven't tried running the lambdas, but they build and deploy.


cdk/src/open-data-platform/data-plane/data-import/write-demographic-data.handler.ts
```diff
@@ -83,255 +75,13 @@ function getTableRowFromRow(row: any): SqlParametersList {
...
-  console.log(`Number errors during insert: ${errors.length}`);
-
-  // Log all the rejected promises to diagnose issues.
-  errors.forEach((error) => console.log(`Insert error: ${error}`));
-}
-
-
 /**
  * Parses S3 'alabama_acs_data.csv' file and writes rows
  * to demographics table in the MainCluster postgres db.
```
kailamjeter, line 80: nit: can you add the total # of rows to this comment since the const DEFAULT_NUMBER_ROWS_TO_INSERT is no longer present?

cdk/src/open-data-platform/data-plane/data-import/write-demographic-data.handler.ts
```diff
@@ -83,255 +75,13 @@ function getTableRowFromRow(row: any): SqlParametersList {
...
-  const errors = result.filter((p) => p.status == 'rejected') as PromiseRejectedResult[];
-  console.log(`Number errors during insert: ${errors.length}`);
-
-  // Log all the rejected promises to diagnose issues.
-  errors.forEach((error) => console.log(`Insert error: ${error}`));
-}
-
-
 /**
  * Parses S3 'alabama_acs_data.csv' file and writes rows
```
kailamjeter, line 79: I know you didn't add this but could you correct this filename?

kailamjeter (approved):

cdk/src/open-data-platform/data-plane/data-import/write-demographic-data.handler.ts
beekley: Done

cdk/src/open-data-platform/data-plane/data-import/write-demographic-data.handler.ts
beekley: How many rows are there? The `DEFAULT_NUMBER_ROWS_TO_INSERT ` was 47841. Is that the total number?

beekley: (I'm ignoring this for now so I can merge this in)

PR #55, opened 2022-06-30
Set up deployment pipeline with dev stage

beekley:
## Description

Currently, we have sandbox environments for each developer. We want to have environments that aren't specific to devs and have (relatively) current code.

This PR sets up a pipeline to deploy a `DEV` environment, which is just like the `SANDBOX` envs, but pulls from the `main` branch of this repo.

### New

- A new CDK app, `pipeline`.
- Added an `envType` variable to the stack, so that each environment has its own namespace.

### Changed

- Lowered the required Node version for the client to 16, since there was a dependency that needed <=17, which is [not supported anymore](https://nodejs.org/en/about/releases/).

## Testing and Reviewing

I could synth this CDK config, but not deploy it as I don't have write access to the Deployments account. The Github connector may need to be given access to this repo as well.


cdk/src/util.ts
```diff
@@ -1,25 +1,35 @@
...
+}
 
 // Each construct in AWS is prefixed with its stack's name. This function creates a useful name
 // for identification of the construct.
-// TODO: drop developer user name from the stack for non-dev environments.
-export const stackName = (id: StackId): string => {
+export const stackName = (id: StackId, e: EnvType): string => {
   const base = `${projectName}${StackId[id]}`;
-  // Share a common Network name across multiple developers, so don't include the user name.
+  // Network stuff should be shared across environments within a given AWS account, so use a
```
kailamjeter, line 30: also between environments?

kailamjeter (approved):

cdk/src/pipeline/pipeline-stack.ts
```diff
@@ -0,0 +1,48 @@
...
+          // Update npm.
+          'npm i -g npm',
+
+          // Build the client.
+          'npm --prefix client ci',
+          'npm --prefix client run build',
+
+          // Synthesize the CDK app.
+          'npm --prefix run cdk ci',
+          'npm --prefix run cdk pipeline-synth',
```
arockett, line 34: I haven't used the ```--prefix``` option before, but wouldn't this need to be

```
'npm --prefix cdk ci',
'npm --prefix cdk run pipeline-synth'
```

cdk/src/pipeline/pipeline-stack.ts
```diff
@@ -0,0 +1,48 @@
...
+        ],
+        primaryOutputDirectory: 'cdk/cdk.out',
+      }),
+    });
+
+    pipeline.addStage(
+      new OpenDataPlatformStage(scope, 'Development', {
```
arockett, line 41: Please us the id 'Dev' so we can keep the stack names short and consistent across projects

cdk/src/open-data-platform/root-stack.ts
```diff
@@ -9,22 +9,27 @@ import { FrontendStack } from './frontend/frontend-stack';
 import { AppPlaneStack } from './app-plane/app-plane-stack';
 
 export class OpenDataPlatformStack extends Stack {
-  constructor(scope: Construct, id: string, props?: util.CommonProps) {
+  constructor(scope: Construct, id: string, props: util.CommonProps) {
     super(scope, id, props);
+    const { envType } = props;
 
-    const networkStack = new NetworkStack(scope, util.stackName(util.StackId.Network), {
+    const networkStack = new NetworkStack(scope, util.stackName(util.StackId.Network, envType), {
```
arockett, line 16: Is 'scope' instead of 'this' used intentionally here and in other places? What is the difference?

arockett (changes requested):
Mostly looks good but I'm wondering how the stack naming utilities plus pipeline stage and nested stack structure you've gone with will effect the end stack names which are deployed through the pipeline.

Since this cdk project's stage and stacks aren't set up exactly the same as the Patina/Tributary pipelines, there may be a bit of a back and forth up front to make sure the pipeline deploys with the names we want. You and I should schedule some time to just knock out that iteration for early next week.

cdk/src/open-data-platform/root-stack.ts
beekley: (as discussed) this was unintentional and meant that each of these stacks aren't actually nested stacks.

I've removed the root stack altogether now.

cdk/src/pipeline/pipeline-stack.ts
beekley: Good catch, thanks.

cdk/src/util.ts
beekley: I've rephrased.

cdk/src/util.ts
```diff
@@ -1,25 +1,35 @@
...
+  // common name for the network stack.
   if (id === StackId.Network) return base;
-  return `${process.env.USER}-${base}`;
+  if (e == EnvType.Sandbox && process.env.USER) return `${process.env.USER}-${base}`;
+  return `${e}-${base}`;
```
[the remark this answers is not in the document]
beekley: Thanks, I've removed that prefix.

arockett (approved):
LGTM, let's schedule time to init the pipeline in the Deployments account and see if everything works as expected

PR #56, opened 2022-06-30
Run the test suite before merging code.

beekley:
## Description

This adds a [Github Action](https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-nodejs) that builds the client and runs the cdk test suite, so we only merge code in that passes tests.

### New

- Github action

## Testing and Reviewing

[It works!](https://github.com/BlueConduit/open-data-platform/runs/7141268596?check_suite_focus=true)


kailamjeter (approved):

client/package.json
```diff
@@ -58,6 +58,6 @@
   ],
   "engines": {
     "npm": ">=8.0.0",
-    "node": ">=17.0.0"
+    "node": ">=16.0.0"
```
justinebreuch, line 61: Curious why this change?

justinebreuch (approved):

client/package.json
beekley: Node 17 isn't supported anymore and a dependency requires Node <=17. So 16 is the most recent, supported version.

(I had to make the same change in my [other PR](https://github.com/BlueConduit/open-data-platform/pull/55) but forgot to call it out here)

PR #57, opened 2022-07-07
Add state and county columns to demographics

justinebreuch:
## Description

Addresses: [Complete data transformation](https://app.shortcut.com/blueconduit/story/5678/complete-data-transformation)

Add state and county geo id to the demographics table. We need this in order to do joins with state and county tables. Geometric spatial joins are too slow for our purposes. 

The table needs to be dropped and reloaded to make all the aggregations work. Aggregations will require these foreign keys be populated.

### New

State and county geo ids added to demographics table. 

## Testing and Reviewing

Reloaded entire demographics table with new columns

### DIFF
--- a/cdk/src/open-data-platform/data-plane/schema/schema.sql
+++ b/cdk/src/open-data-platform/data-plane/schema/schema.sql
@@ -37,9 +37,9 @@ CREATE TABLE IF NOT EXISTS demographics(
    under_five_population real,
    poverty_population real,
    black_population real,
+   white_population real,
    census_state_geo_id varchar(255) NOT NULL references states(census_geo_id),
    census_county_geo_id varchar(255) NOT NULL references counties(census_geo_id),
-   white_population real,
    geom GEOMETRY(Geometry, 4326),
    PRIMARY KEY(census_geo_id)
 );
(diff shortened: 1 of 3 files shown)


cdk/src/open-data-platform/data-plane/schema/schema.sql
```diff
@@ -31,16 +31,20 @@ $$ LANGUAGE plpgsql;
...
-    white_population real,
-    geom GEOMETRY(Geometry, 4326),
-    PRIMARY KEY(census_geo_id)
+   census_geo_id varchar(255) NOT NULL,
+   census_block_name varchar(255),
+   total_population real,
+   under_five_population real,
+   poverty_population real,
+   black_population real,
+   census_state_geo_id varchar(255) NOT NULL references states(census_geo_id),
```
beekley, line 40: Move to an `ALTER TABLE` statement.

beekley (changes requested):

cdk/src/open-data-platform/data-plane/schema/schema.sql
justinebreuch: None of us has a finished table yet. This table will need to be dropped even after this PR to get the zipcode in, unfortunately. I'll be sure to do ALTER TABLE statements once we have finalized tables / are not just writing with lambdas

beekley: Can you describe that in the PR description then? Otherwise I assume I wouldn't need to drop a table to get a schema update.

beekley (approved):

cdk/src/open-data-platform/data-plane/schema/schema.sql
justinebreuch: Added a note about dropping to make aggregations work

PR #64, opened 2022-07-13
UI updates for 0.1

justinebreuch:
## Description

Addresses: [Complete data transformation](https://app.shortcut.com/blueconduit/story/5678/complete-data-transformation)

Refactors some of the MapView in order to fix the routing bug where the first load won't work when a param is passed in.
- Now the currentState does not need to be set by the App. It can be set by the MapView. 

Updates demographics data layer to toggle between states / county. 
- The legend does not yet update. I added a TODO. Mapbox expressions are tricky so I need to take another pass at it. 

Adds formatting to the popups based on data type / refactors how buckets work 

## Testing and Reviewing

UI


client/src/components/MapPopupContent.vue
```diff
@@ -67,13 +67,17 @@ export default defineComponent({
...
           case FeaturePropertyDataType.String: {
             propertyValue = propertyValue ?? '';
             break;
           }
           case FeaturePropertyDataType.Number: {
-            propertyValue = parseFloat(propertyValue ?? '0').toLocaleString() ;
+            propertyValue = parseFloat(propertyValue ?? '0').toLocaleString();
+            break;
+          }
+          case FeaturePropertyDataType.Percent: {
```
kailamjeter, line 79: Percentage? Feel like data type -> percentage is more correct (same below)

client/src/components/MapView.vue
```diff
@@ -93,10 +94,10 @@ export default defineComponent({
         this.popup.remove();
       }
 
-      if (newDataLayer != null && !this.isDataLayerVisible(newDataLayer.styleLayer.id)) {
+      if (newDataLayer != null) {
         this.setDataLayerVisibility(newDataLayer.styleLayer.id, true);
       }
-      if (oldDataLayer != null && this.isDataLayerVisible(oldDataLayer.styleLayer.id)) {
```
kailamjeter, line 99: is this function still used (isDataLayerVisible)? if not can we remove?

client/src/components/MapView.vue
```diff
@@ -208,6 +207,13 @@ export default defineComponent({
 
       this.setUpInteractionHandlers();
       this.setUpControls();
+
+      // Check whether there's a layer selected in the router.
+      if (router.currentRoute.value.query?.layer != null) {
+        this.state.currentDataLayer = this.state.dataLayers.find(l => l.styleLayer.id == router.currentRoute.value.query.layer) ?? leadServiceLinesByWaterSystemLayer;
```
kailamjeter, line 213: nit: setCurrentDataLayer()

kailamjeter, line 213: nit: layer instead of 'l' - it looks like a 1. Also can we not just check whether the query param is present? if the param value is invalid shouldn't we remove it instead of ignoring it?

client/src/data_layer_configs/population_by_census_block_config.ts
```diff
@@ -7,46 +7,93 @@ import {
...
+  },
+  {
+    bucketValue: ONE_MILLION * 20,
+    bucketColor: '#0B2553',
+  },
 ];
 
+// TODO (breuch): automate this. Right now the mathematical relationship
+// between LEGEND_VALUES_STATES and LEGEND_VALUES_COUNTY is not yet clear.
+export const LEGEND_VALUES_COUNTY = [
```
kailamjeter, line 46: thanks for cleaning up these look great!

client/src/data_layer_configs/population_by_census_block_config.ts
```diff
@@ -72,38 +122,32 @@ export const styleLayer: FillLayer = {
...
-      label: 'Total population',
+      label: 'Total residents',
       name: 'total_population',
       dataType: FeaturePropertyDataType.Number,
     },
     {
-      label: 'Population under the age of 5',
+      label: 'Residents under the age of 5',
       name: 'under_five_population',
       dataType: FeaturePropertyDataType.Number,
```
kailamjeter, line 137: should these be percentages? as in can we convert them in the FE to %.

client/src/util/data_layer_util.ts
```diff
@@ -1,20 +1,46 @@
...
+    switch (legend.bucketLabelType) {
+      case FeaturePropertyDataType.String: {
+        bucket.bucketLabel = bucket.bucketValue ?? '';
+        break;
+      }
+      case FeaturePropertyDataType.Number: {
+        bucket.bucketLabel = parseFloat(bucket.bucketValue ?? '0').toLocaleString();
+        break;
+      }
+      case FeaturePropertyDataType.Percent: {
```
kailamjeter, line 18: what field comes to FE as % already?

client/src/components/MapView.vue
justinebreuch: Thank you for this suggestion! Just saw it on your PR

client/src/data_layer_configs/population_by_census_block_config.ts
justinebreuch: I need to figure out how to generate one legend value from another without having two maps

client/src/components/MapView.vue
justinebreuch: You point out something good because I realize I can just do 
this.state.setCurrentDataLayer(this.state.dataLayers.find(layer => layer.styleLayer.id == router.currentRoute.value.query?.layer)
        ?? leadServiceLinesByWaterSystemLayer);

Now if there is a misspelling in the routing it just redirects to the water systems data layer

client/src/util/data_layer_util.ts
justinebreuch: It's not from the DB it's from the interpolation https://github.com/BlueConduit/open-data-platform/pull/64/files#diff-8e016a7a6487cb36ceffea3535fbc4d7e4a70f3b73419ff3dc2e8b768472b087

client/src/components/MapView.vue
justinebreuch: Nope! Removed

client/src/data_layer_configs/population_by_census_block_config.ts
justinebreuch: Added a todo! Definitely agree that's easier to read

client/src/util/data_layer_util.ts
kailamjeter: ohh nice - I added logic to do *100 for percentages in the handler to parse the property values - how do you feel about that vs this? I might be misunderstanding, but does modifying that value in order to color the map also change the value in the feature property list?

kailamjeter (approved):

PR #67, opened 2022-07-13
Add DNS

beekley:
## Description

This PR copies the Tributary DNS config so each environment uses its own domain name instead of the default cloudfront one. I'm not committed to this subdomain, but wanted to get something in there for now.

### New

- DNS config.

## Testing and Reviewing

Added a basic test that CloudFront receives the DNS config. Also my sandbox now has a URL (though won't load for other reasons): https://beekley.sandbox.leadout.blueconduit.com/


cdk/src/open-data-platform/network/dns.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    const { envType } = props;
+
+    const parentDomain = 'blueconduit.com';
+
+    let subdomain = `default.${baseSubdomain}`;
+    if (envType === EnvType.Production) subdomain = baseSubdomain;
+    else if (envType === EnvType.Development) subdomain = `dev.${baseSubdomain}`;
+    else if (envType === EnvType.Sandbox)
+      subdomain = `${process.env.USER ?? 'default'}.sandbox.${baseSubdomain}`;
```
arockett, line 26: I don't think this subdomain structure for dev and sandboxes will work with domain delegation. Except for the root 'blueconduit.com' domain, there can be no overlap between env domains.

Example:

If the root 'blueconduit.com' hosted zone delegates DNS for the 'leadout' subdomain to the prod environment, then even if there's a record that tries to delegate the 'dev.leadout' subdomain to a different env, it will still go to the prod environment

arockett: We handle this in the saas app because prod is 'app.blueconduit.com', dev is 'dev.blueconduit.com' and our sandboxes are all unique at '${user}.sandbox.blueconduit.com' AND there is no 'sandbox.blueconduit.com' environment

arockett (changes requested):
Review the 'leadout' subdomain name with your team and our team before merging and revisit the subdomains you use for different environments.

cdk/src/open-data-platform/network/dns.ts
arockett: proposal for this app is to use something like 'leadout-dev' as the subdomain for the dev env and '${user}.leadout-sandbox' for sandbox envs

cdk/src/util.ts
```diff
@@ -16,6 +16,7 @@ export enum EnvType {
 }
 export const defaultEnv = EnvType.Sandbox;
 export const projectName = 'OpenDataPlatform';
+export const baseSubdomain = 'leadout';
```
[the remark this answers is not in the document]
beekley: [The team has spoken](https://lslr.slack.com/archives/C03D8ECRC3S/p1658179723576249)

cdk/src/open-data-platform/network/dns.ts
beekley: Thanks for that info. I've updated to follow that pattern.

arockett (approved):
LGTM!

PR #73, opened 2022-07-21
Add initial tables needed for scorecard. Posting for testing on laptop.

kailamjeter:
## Description

Add tables to hold scorecard data.  

See: https://docs.google.com/spreadsheets/d/1aJV4BP5NyZM7Hrp682AMQbD9IEzt0JmzNVcVczf4Zj0

### New

- Tables for lead predictions and U.S. demographic data needed for scorecard

### Changed

- schema.sql

### Removed

- N/A

## Testing and Reviewing

paste new code into query editor

### DIFF
--- a/cdk/src/open-data-platform/data-plane/data-import/write-water-systems-data.handler.ts
+++ b/cdk/src/open-data-platform/data-plane/data-import/write-water-systems-data.handler.ts
@@ -28,13 +28,15 @@ async function insertBatch(
     secretArn: process.env.CREDENTIALS_SECRET ?? '',
     sql: `INSERT INTO water_systems (pws_id,
                                      pws_name,
-                                     lead_connections_count,
+                                     lead_connections_low_estimate,
+                                     lead_connections_high_estimate,
                                      service_connections_count,
                                      population_served,
                                      geom)
           VALUES (:pws_id,
                   :pws_name,
-                  :lead_connections_count,
+                  :lead_connections_low_estimate,
+                  :lead_connections_high_estimate,
                   :service_connections_count,
                   :population_served,
                   ST_AsText(ST_GeomFromGeoJSON(:geom))) ON CONFLICT (pws_id) DO NOTHING`,
@@ -60,7 +62,9 @@ function getTableRowFromRow(row: any): SqlParametersList {
     new WaterSystemsTableRowBuilder()
       .pwsId(properties.pwsid)
       .pwsName(properties.pws_name ?? '')
-      .leadConnectionsCount(getValueOrDefault(properties.lead_connections))
+      // TODO(kailamjeter): modify property name when column names are modified.
+      .leadConnectionsLowEstimate(getValueOrDefault(properties.lead_connections))
+      .leadConnectionsHighEstimate(getValueOrDefault(properties.lead_connections))
       .serviceConnectionsCount(getValueOrDefault(properties.service_connections_count))
       .populationServed(getValueOrDefault(properties.population_served_count))
       // Keep JSON formatting. Post-GIS helpers depend on this.
--- a/cdk/src/open-data-platform/data-plane/data-import/write-water-systems-data.handler.ts
+++ b/cdk/src/open-data-platform/data-plane/data-import/write-water-systems-data.handler.ts
@@ -28,15 +28,13 @@ async function insertBatch(
     secretArn: process.env.CREDENTIALS_SECRET ?? '',
     sql: `INSERT INTO water_systems (pws_id,
                                      pws_name,
-                                     lead_connections_low_estimate,
-                                     lead_connections_high_estimate,
+                                     lead_connections_count,
                                      service_connections_count,
                                      population_served,
                                      geom)
           VALUES (:pws_id,
                   :pws_name,
-                  :lead_connections_low_estimate,
-                  :lead_connections_high_estimate,
+                  :lead_connections_count,
                   :service_connections_count,
                   :population_served,
                   ST_AsText(ST_GeomFromGeoJSON(:geom))) ON CONFLICT (pws_id) DO NOTHING`,
@@ -62,9 +60,7 @@ function getTableRowFromRow(row: any): SqlParametersList {
     new WaterSystemsTableRowBuilder()
       .pwsId(properties.pwsid)
       .pwsName(properties.pws_name ?? '')
-      // TODO(kailamjeter): modify property name when column names are modified.
-      .leadConnectionsLowEstimate(getValueOrDefault(properties.lead_connections))
-      .leadConnectionsHighEstimate(getValueOrDefault(properties.lead_connections))
+      .leadConnectionsCount(getValueOrDefault(properties.lead_connections))
       .serviceConnectionsCount(getValueOrDefault(properties.service_connections_count))
       .populationServed(getValueOrDefault(properties.population_served_count))
       // Keep JSON formatting. Post-GIS helpers depend on this.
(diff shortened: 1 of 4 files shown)


cdk/src/open-data-platform/data-plane/schema/schema.sql
```diff
@@ -95,22 +95,44 @@ CREATE INDEX IF NOT EXISTS census_state_geo_id_index ON demographics (state_cens
 CREATE INDEX IF NOT EXISTS census_county_geo_id_index ON demographics (county_census_geo_id);
 CREATE INDEX IF NOT EXISTS geom_index ON demographics USING GIST (geom);
 
+--- Tables related to U.S. demographic information.
+
+-- TODO(breuch): Add insert statement once the zipcode -> demographics
+
+CREATE TABLE IF NOT EXISTS aggregate_us_demographics
```
beekley, line 102: What does each row in this table represent? I'd rather name the table after that (e.g. `parcels` is a table of parcels).

cdk/src/open-data-platform/data-plane/schema/schema.sql
```diff
@@ -173,92 +197,21 @@ EXECUTE PROCEDURE update_last_update_timestamp();
...
-    under_five_population real,
-    poverty_population    real,
-    geom                  geometry(Geometry, 3857),
-    PRIMARY KEY (census_geo_id)
-);
-CREATE INDEX IF NOT EXISTS geom_index ON zipcode_demographics USING GIST (geom);
-
 -- Lead connections aggregation tables.
 
 CREATE TABLE IF NOT EXISTS state_lead_connections
```
beekley, line 202: Is there a reason this isn't in the `states` table? That would make it behave more like other tables like `parcels`.

beekley (changes requested):
LGTM, with two naming/organizational questions.

cdk/src/open-data-platform/data-plane/schema/schema.sql
kailamjeter: With this schema the rows would represent either a zipcode, county, or state (indicated in type column with a string). Each has a geo ID which allows all of them to be unified in a single table. If we would rather organize by the geographic type I could make an individual table for each one, however it doesn't seem necessary here IMO.

cdk/src/open-data-platform/data-plane/schema/schema.sql
kailamjeter: Good point - I didn't create this table in this PR, it was created when I created aggregate tables for lead connections (see deleted tables for counties + zips below). At that point it was because we generally were separating out the aggregated data into different tables that were only being used for function sources, and state / county / zip tables only stored information directly from the cartographic boundaries source. Agreed that it makes sense to unify these metrics into the state table now that we aren't storing by source.

kailamjeter: As discussed offline this will require a good amount of refactoring to pull from multiple sources. Left a TODO and will document as p1 on eng tasks if it isn't tracked yet.

cdk/src/open-data-platform/data-plane/data-import/write-water-systems-data.handler.ts
```diff
@@ -28,13 +28,15 @@ async function insertBatch(
     secretArn: process.env.CREDENTIALS_SECRET ?? '',
     sql: `INSERT INTO water_systems (pws_id,
                                      pws_name,
-                                     lead_connections_count,
+                                     lead_connections_low_estimate,
```
justinebreuch, line 31: I actually don't think we are getting this anymore. Just lead lines and service lines

cdk/src/open-data-platform/data-plane/schema/schema.sql
```diff
@@ -95,22 +95,44 @@ CREATE INDEX IF NOT EXISTS census_state_geo_id_index ON demographics (state_cens
...
+
+CREATE TABLE IF NOT EXISTS aggregate_us_demographics
+(
+    census_geo_id                varchar(255) NOT NULL,
+    geo_type                     varchar(255) NOT NULL,
+    name                         varchar(255) NOT NULL,
+    average_home_age             real,
+    average_income_level         real,
+    average_social_vulnerability real,
+    black_population             real,
```
justinebreuch, line 110: I don't think you need lines 110-115

cdk/src/open-data-platform/data-plane/schema/schema.sql
```diff
@@ -95,22 +95,44 @@ CREATE INDEX IF NOT EXISTS census_state_geo_id_index ON demographics (state_cens
...
-    lead_connections_count    real,
-    pws_name                  varchar(255),
-    service_connections_count real,
-    population_served         real,
-    state_census_geo_id       varchar(255) references states (census_geo_id),
-    county_census_geo_id      varchar(255) references counties (census_geo_id),
-    geom                      GEOMETRY(Geometry, 4326),
+    pws_id                         varchar(255) NOT NULL,
+    pws_name                       varchar(255),
+    lead_connections_low_estimate  real,
```
justinebreuch, line 127: Same here I don't think we get it but we can verify with Jared

cdk/src/open-data-platform/data-plane/schema/schema.sql
```diff
@@ -173,92 +197,22 @@ EXECUTE PROCEDURE update_last_update_timestamp();
...
-    under_five_population real,
-    poverty_population    real,
-    geom                  geometry(Geometry, 3857),
-    PRIMARY KEY (census_geo_id)
-);
-CREATE INDEX IF NOT EXISTS geom_index ON county_demographics USING GIST (geom);
-
--- TODO(breuch): Add insert statement once the zipcode -> demographics
--- connection established.
-CREATE TABLE IF NOT EXISTS zipcode_demographics
```
justinebreuch, line 208: This would break the API endpoint checked in today. Can you update the zip_codes/{zip_code}/scorecard API to reference the aggregated table instead of zipcode_demos?

cdk/src/open-data-platform/data-plane/schema/schema.sql
beekley: sounds good!

cdk/src/open-data-platform/data-plane/data-import/write-water-systems-data.handler.ts
beekley: Hmm. Should we wait to create schema changes until we have the data and know what it looks like?

beekley (approved):

cdk/src/open-data-platform/data-plane/data-import/write-water-systems-data.handler.ts
kailamjeter: Is this documented Justine? I can hold off submitting until we get sources but it will block e2e testing if we are using the "new" schema in API.

cdk/src/open-data-platform/data-plane/schema/schema.sql
kailamjeter: SG - wasn't sure of best way to migrate over while we don't have no data, but I'll assume we want this to look as close to end state as possible.

cdk/src/open-data-platform/data-plane/schema/schema.sql
kailamjeter: Done

cdk/src/open-data-platform/data-plane/schema/schema.sql
kailamjeter: Done.

cdk/src/open-data-platform/data-plane/data-import/write-water-systems-data.handler.ts
kailamjeter: Updated but still awaiting Jared's response.

justinebreuch (approved):

PR #74, opened 2022-07-25
Add /geolocate endpoint that returns all geoidentifers for a given lat,long 

justinebreuch:
## Description

Addresses: API skeleton (AWS api gateway + lambda creation) C1_FR2

Adds lambda and api gateway endpoint for /geolocate which queries our geometry tables to find intersections with a lat,long pair. Reflects design in [here](https://docs.google.com/document/d/1T-hBIzci81r0Xez70zEVfYxWHh6oHKsXJlxDsVJ7lDM/edit?usp=sharing)

### New

ApiStack which initializes the gateway resource and endpoint
Geolocate lambda 

## Testing and Reviewing

Tested via AWS Gateway test event


justinebreuch:
Feel free to ignore the review, Alex. I noticed Aaron was out and you're a reviewer on the doc so I thought I'd loop you in

cdk/src/open-data-platform/data-plane/api/api-stack.ts
```diff
@@ -0,0 +1,64 @@
...
+import * as path from 'path';
+
+export class ApiStack extends Construct {
+  constructor(scope: Construct, id: string, props: SchemaProps) {
+    super(scope, id);
+
+    const { cluster, vpc, db, credentialsSecret } = props;
+
+    const lambdaFunction = new lambda.NodejsFunction(this, 'geolocate-handler', {
+      entry: `${path.resolve(__dirname + '/../../../../../')}/api/src/geolocate/get.handler.ts`,
```
justinebreuch, line 17: It feels like there should be a better way to do this but the absolute path wouldn't work here

cdk/src/open-data-platform/data-plane/api/api-stack.ts
```diff
@@ -0,0 +1,64 @@
...
+      deployOptions: {
+        stageName: 'dev',
+      },
+
+      // Enable CORS
+      defaultCorsPreflightOptions: {
+        allowHeaders: ['Content-Type', 'X-Amz-Date', 'Authorization', 'X-Api-Key'],
+        allowMethods: ['GET'],
+        allowCredentials: true,
+        allowOrigins: ['http://localhost:3000'],
```
justinebreuch, line 57: What domains need to be added here? Sandboxes and then eventually dev / prod?

api/src/geolocate/get.handler.ts
```diff
@@ -0,0 +1,95 @@
...
+          WHERE geom && ST_SetSRID(ST_Point(:long, :lat), 4326)
+          LIMIT 1`,
+    parameters: params,
+  };
+  const results = await rdsService.executeStatement(executeParams).promise();
+  const geoids = results.records?.map((record: FieldList) => record[0].stringValue) ?? [];
+  return geoids[0];
+}
+
+export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
```
beekley, line 41: The `APIGatewayProxyEvent` is broad and doesn't include our param types. Let's define the request interface here, similar to `GeolocateApiResponse`. E.g.
```
interface GeolocateApiRequest {
  lat: number;
  long: number;
}
```

cdk/src/open-data-platform/data-plane/api/api-stack.ts
beekley: https://nodejs.org/api/process.html#processcwd might help here.

cdk/src/open-data-platform/data-plane/api/api-stack.ts
beekley: Can this be a regex? If so, using a wildcard subdomain like `https?//.*\.blueconduit\.com/` would be ok.

If not, we could get the hostname [from DNS directly](https://github.com/BlueConduit/open-data-platform/blob/main/cdk/src/open-data-platform/frontend/frontend-stack.ts#L41).

beekley (changes requested):
LGTM overall. Small changes requested.

api/src/geolocate/get.handler.ts
```diff
@@ -0,0 +1,95 @@
...
+): Promise<string | undefined> {
+  const executeParams: ExecuteStatementRequest = {
+    database: process.env.DATABASE_NAME ?? 'postgres',
+    resourceArn: process.env.RESOURCE_ARN ?? '',
+    schema: SCHEMA,
+    secretArn: process.env.CREDENTIALS_SECRET ?? '',
+    sql: `SELECT ${geoid}
+          FROM ${table}
+          WHERE geom && ST_SetSRID(ST_Point(:long, :lat), 4326)
+          LIMIT 1`,
```
beekley, line 33: Will there ever be more than one shape that intersects a given point? If so, I'm concerned about throwing away that data by limiting this to the first result. If not, we should update the design doc, which says there can be multiple water systems returned for the /geolocate endpoint.

api/src/geolocate/get.handler.ts
```diff
@@ -0,0 +1,95 @@
...
+  table: string,
+): Promise<string | undefined> {
+  const executeParams: ExecuteStatementRequest = {
+    database: process.env.DATABASE_NAME ?? 'postgres',
+    resourceArn: process.env.RESOURCE_ARN ?? '',
+    schema: SCHEMA,
+    secretArn: process.env.CREDENTIALS_SECRET ?? '',
+    sql: `SELECT ${geoid}
+          FROM ${table}
+          WHERE geom && ST_SetSRID(ST_Point(:long, :lat), 4326)
```
beekley, line 32: This might need to use [`ST_Contains`](https://postgis.net/docs/ST_Contains.html) to check that the point is within `geom` ([possibly helpful example](https://gis.stackexchange.com/a/83555).

cdk/src/open-data-platform/data-plane/api/api-stack.ts
```diff
@@ -0,0 +1,64 @@
...
+    if (lambdaFunction.role?.roleArn != undefined) {
+      let role = iam.Role.fromRoleArn(
+        scope,
+        id + '-' + lambdaFunction,
+        lambdaFunction.role.roleArn,
+      );
+      cluster.grantDataApiAccess(role);
+    }
+
+    const api = new apigateway.RestApi(this, 'open-data-platform-api', {
```
beekley, line 46: [Can you use this](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda.FunctionUrlOptions.html) to create a URL without setting up API Gateway?

cdk/src/open-data-platform/data-plane/api/api-stack.ts
justinebreuch: Thank you!

api/src/geolocate/get.handler.ts
justinebreuch: The path params are in the event so I can't change the argument to the handler

api/src/geolocate/get.handler.ts
justinebreuch: Added a todo. Once we search by county, state that will be true (multiple water systems)

api/src/geolocate/get.handler.ts
justinebreuch: Added a todo for sake of getting this in

cdk/src/open-data-platform/data-plane/api/api-stack.ts
justinebreuch: Decided offline: going to check this in to unblock FE work. Brett's going to look into converting from API gateway --> cloudfront to get a domain for this api instead of creating a new domain

cdk/src/open-data-platform/data-plane/api/api-stack.ts
```diff
@@ -14,7 +15,8 @@ export class ApiStack extends Construct {
     const { cluster, vpc, db, credentialsSecret } = props;
 
     const lambdaFunction = new lambda.NodejsFunction(this, 'geolocate-handler', {
-      entry: `${path.resolve(__dirname + '/../../../../../')}/api/src/geolocate/get.handler.ts`,
+      //entry: `${path.resolve(__dirname + '/../../../../../')}/api/src/geolocate/get.handler.ts`,
```
beekley, line 18: Remove this comment.

beekley (changes requested):

api/src/geolocate/get.handler.ts
justinebreuch: It's intersects in postgis https://postgis.net/docs/geometry_overlaps.html

cdk/src/open-data-platform/data-plane/api/api-stack.ts
justinebreuch: Done

api/src/geolocate/get.handler.ts
justinebreuch: Removed the todo as well

cdk/package.json
```diff
@@ -39,6 +39,7 @@
     "moment": "^2.29.3",
     "pg": "^8.7.3",
     "pg-format": "^1.0.4",
+    "process": "^0.11.10",
```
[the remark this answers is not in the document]
justinebreuch: I was getting: /Users/breuch/BlueConduit/open-data-platform/cdk/node_modules/ts-node/src/index.ts:513
    return new TSError(diagnosticText, diagnosticCodes)
           ^
TSError: ⨯ Unable to compile TypeScript:
src/open-data-platform/data-plane/api/api-stack.ts:8:21 - error TS2307: Cannot find module 'node:process' or its corresponding type declarations.

8 import { cwd } from 'node:process';
                      ~~~~~~~~~~~~~~
when I tried it originally and it seems to be due to the older version of node we are using but I remember you did that on purpose for a cdk reason? https://github.com/firebase/firebase-tools/issues/4595

api/src/geolocate/get.handler.ts
justinebreuch: I didn't realize you could override that way. Updated

cdk/package.json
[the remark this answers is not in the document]
beekley: Ah, looks like Node 18 uses `node:process` but below that uses `process`. Since CodeBuild doesn't support Node 18, let's use Node 16 on our local machines to match CodeBuild. That should allow you to import from `process` without installing a 3rd party package.

Then we can define that constraint in package.json, similar to [how we do for the client](https://github.com/BlueConduit/open-data-platform/blob/main/client/package.json#L62).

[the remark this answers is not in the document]
justinebreuch: Ah thank you! That worked!

cdk/package.json
```diff
@@ -42,5 +42,9 @@
     "source-map-support": "^0.5.16",
     "stream-chain": "^2.2.5",
     "stream-json": "^1.7.4"
+  },
+  "engines": {
+    "npm": ">=8.0.0",
+    "node": ">=16.0.0"
```
beekley, line 48: This should be <18, since 18 introduces breaking changes that make importing `process` fail.

justinebreuch: Done!

beekley (approved):

PR #75, opened 2022-07-25
Brettbeekley/sc 5985/add mapbox access token to dev

beekley:
## Description

Previously, the client used a Mapbox access token that each dev had in their local env variables. Since those env variables isn't available to AWS CodeBuild during deployment, this PR checks in a config file that sets them.

The access token in this file was created just for LeadOut and is restricted to `*.blueconduit.com` domains, per [the production design doc](https://docs.google.com/document/d/1zZxCoXx5JzLXTOGVdvC4s8H-r82DFjfmNU-dmFPAQVI/edit#heading=h.a5e8qo5ga0ap)

### Changed

- Moved the access token to a config file.
- Some linting stuff, since the linter was complaining as I worked with this variable.

## Testing and Reviewing

Deployed to my sandbox successfully. I checked that this new token is being used by the client in Chrome dev tools.

### DIFF
--- a/client/src/components/MapGeocoderWrapper.vue
+++ b/client/src/components/MapGeocoderWrapper.vue
@@ -49,7 +49,6 @@ export default defineComponent({
       // app load since the map is only created once.
       if (newMap != null) {
         this.geocoder = new MapboxGeocoder({
-          // TODO(kailamjeter): use secrets manager for this.
           accessToken: process.env.VUE_APP_MAP_BOX_API_TOKEN ?? '',
           mapboxgl: undefined,
           marker: false,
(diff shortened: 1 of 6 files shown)


justinebreuch (approved):

client/src/components/MapGeocoderWrapper.vue
```diff
@@ -50,7 +50,7 @@ export default defineComponent({
       if (newMap != null) {
         this.geocoder = new MapboxGeocoder({
           // TODO(kailamjeter): use secrets manager for this.
```
justinebreuch, line 52: I think this comment can go now

beekley: Done.

PR #76, opened 2022-07-27
Call /geolocate/latlng when user searches zipcode 

justinebreuch:
## Description

Addresses: Implement API calls to use url params (lat/long) to fetch scorecard data C1_FR2

Adds search bar that calls /geolocate/latlng based on input and console.log()s the output. This only happens on "/" not "/map". The next PR will update the app state with the lat long and move the API request out of the geocoder

### Removed

- The style we were using was broken in mapbox studio so I'm using another for now

## Testing and Reviewing

Searched my zipcode and it returns the API response

### DIFF
--- a/client/src/components/MapContentContainer.vue
+++ b/client/src/components/MapContentContainer.vue
@@ -23,8 +23,8 @@ export default defineComponent({
   },
   data() {
     return {
-      showSearch: router.currentRoute.value.path.includes(MAP_ROUTE),
-      showResult: router.currentRoute.value.path.includes(HOME_ROUTE),
+      showSearch: router.currentRoute.value.path = MAP_ROUTE,
+      showResult: router.currentRoute.value.path == HOME_ROUTE,
     };
   },
 });
--- a/client/src/components/MapContentContainer.vue
+++ b/client/src/components/MapContentContainer.vue
@@ -23,7 +23,7 @@ export default defineComponent({
   },
   data() {
     return {
-      showSearch: router.currentRoute.value.path = MAP_ROUTE,
+      showSearch: router.currentRoute.value.path == MAP_ROUTE,
       showResult: router.currentRoute.value.path == HOME_ROUTE,
     };
   },
--- a/client/src/components/MapContentContainer.vue
+++ b/client/src/components/MapContentContainer.vue
@@ -1,12 +1,15 @@
 <template>
-  <SearchBar/>
-  <MapContainer/>
+  <SearchBar v-if='showSearch' />
+  <PredictionPanel v-if='showResult' />
+  <MapContainer />
 </template>
 
 <script lang='ts'>
 import MapContainer from '@/components/MapContainer.vue';
 import SearchBar from '@/components/SearchBar.vue';
 import { defineComponent } from 'vue';
+import PredictionPanel from './PredictionPanel.vue';
+import { router, HOME_ROUTE, MAP_ROUTE } from '../router';
 
 /**
  * Container for SearchBar and MapContainer.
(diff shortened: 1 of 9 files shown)


client/src/components/MapContentContainer.vue
```diff
@@ -15,8 +18,15 @@ export default defineComponent({
   name: 'MapContentContainer',
   components: {
     MapContainer,
+    PredictionPanel,
     SearchBar,
   },
+  data() {
+    return {
+      showSearch: router.currentRoute.value.path = MAP_ROUTE,
```
beekley, line 26: should this be `==`?

client/src/components/MapGeocoderWrapper.vue
```diff
@@ -29,20 +30,27 @@ import mapboxgl from 'mapbox-gl';
  */
 export default defineComponent({
   name: 'MapGeocoderWrapper',
-  setup() {
+  setup(props, { emit }) {
+    const visible = computed({
```
beekley, line 34: What does this function call do?

client/src/components/MapGeocoderWrapper.vue
```diff
@@ -29,20 +30,27 @@ import mapboxgl from 'mapbox-gl';
...
+      get: () => props.modelValue,
+      set: (value) => emit('update:modelValue', value),
+    });
     const state: State = inject(stateKey, State.default());
 
     return {
       state,
+      visible,
```
beekley, line 42: Can we call this something like `expanded`? I was initially confused since it doesn't literally describe the visibility of the component.

client/src/components/MapGeocoderWrapper.vue
```diff
@@ -29,20 +30,27 @@ import mapboxgl from 'mapbox-gl';
...
   },
   data() {
     return {
       // Default to empty geocoder instance that is not connected to map or view.
       geocoder: new MapboxGeocoder(),
-      visible: false,
     };
   },
+  props: {
+    'modelValue': { type: Boolean, default: false },
```
beekley, line 52: What does `modelValue` mean? The name makes me think it'd be a value from the model directory, but it's a bool.

client/src/components/PredictionPanel.vue
```diff
@@ -0,0 +1,41 @@
+<template>
+  <div class='container'>
+    <map-geocoder-wrapper v-if='showSearch' v-model='expandSearch' />
```
beekley, line 3: I'd rather remove the search from this component. We already have a search component parallel to this one and removing it from here would tighten its scope to just displaying predictions/scorecard(s).

beekley (changes requested):

client/src/components/MapContentContainer.vue
justinebreuch: Yeah updated

client/src/components/MapGeocoderWrapper.vue
justinebreuch: yeah for sure

client/src/components/MapGeocoderWrapper.vue
justinebreuch: Added a comment but essentially binds the prop value of isExpanded to the v-model value expandSearch. V-model is just a two way binding. I put it in the template instead which may make it more straight forward

client/src/components/MapGeocoderWrapper.vue
justinebreuch: It's the v-model for this component. I did a custom name for clarity

client/src/components/PredictionPanel.vue
justinebreuch: Yeah this is the search bar that needs to be moved to the landing page. If you want me to take it out of here now that you started the landing page, just let me know. I can also leave it in if this PR is going in first for you to copy / paste over?

client/src/components/MapGeocoderWrapper.vue
beekley: Sounds good, thanks.

client/src/components/PredictionPanel.vue
beekley: Sounds good. I'll move it around as necessary in my PR so it's fine as-is here.

beekley (approved):

PR #78, opened 2022-07-28
Support vue-routing in Cloudfront.

beekley:
## Description

(this was part of a a broader PR that was getting too big, so I split it up.

Before, only the `/` route went to the web app (`index.html`). This tells CloudFront to send every URL that would otherwise 403 or 404 (i.e. any URL that isn't a tileserver or API URL) to go to `index.html` as well.

### Changed

- 404/403 redirects.
- Sets up dumb caching in CloudFront too.

## Testing and Reviewing

You can go to specific routes like https://beekley.leadout-sandbox.blueconduit.com/map, while tileserver stuff says in the tileserver: https://beekley.leadout-sandbox.blueconduit.com/tiles/v1/index.json or https://beekley.leadout-sandbox.blueconduit.com/tiles/v1/not-real


cdk/src/open-data-platform/frontend/frontend-stack.ts
```diff
@@ -48,6 +48,20 @@ export class FrontendStack extends Stack {
...
+          httpStatus: 403,
+          responseHttpStatus: 200,
+          responsePagePath: '/index.html',
+        },
+        {
+          httpStatus: 404,
+          responseHttpStatus: 200,
+          responsePagePath: '/index.html',
+        },
+      ],
```
arockett, line 64: What would happen if the tile server threw a 404? Is there a way to override this setting for the /tiles/* path?

arockett (commented):
Looks good, just one question before I approve.

arockett (approved):
LGTM. Dang though, was hoping there would be behavior-level config for error pages.

PR #79, opened 2022-07-28
Add title metadata based on route.

beekley:
## Description

Before, we just used the default `app` title. This uses a dynamic one based on route.

### New

- Added a watcher for route changes that changes the HTML title.

## Testing and Reviewing

You can see the titles changes on https://beekley.leadout-sandbox.blueconduit.com/ and other URLs.

### DIFF
--- a/client/src/App.vue
+++ b/client/src/App.vue
@@ -2,8 +2,9 @@
   <NavigationBar />
 </template>
 
-<script lang='ts'>
+<script lang="ts">
 import { defineComponent, provide, reactive } from 'vue';
+import { RouteLocation } from 'vue-router';
 import '@blueconduit/copper/dist/css/copper.css';
 import NavigationBar from './components/NavigationBar.vue';
 import { State } from './model/state';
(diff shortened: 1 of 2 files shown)


client/public/index.html
```diff
@@ -11,11 +11,11 @@
     <script
       src='https://api.mapbox.com/mapbox-gl-js/v2.8.2/mapbox-gl.js'></script>
     <link rel='icon' href='<%= BASE_URL %>favicon.ico'>
-    <title><%= htmlWebpackPlugin.options.title %></title>
+    <title>LeadOut</title>
```
justinebreuch, line 14: Do you want to move this to assets/messages/common_messages.ts so we can start moving text out?

client/src/App.vue
```diff
@@ -41,9 +43,15 @@ export default defineComponent({
       state,
     };
   },
+  watch: {
+    $route(to: RouteLocation) {
+      // TODO: consider adding a string that says this is a non-prod environment, so devs can see
```
justinebreuch, line 48: How come?

client/public/index.html
beekley: That's a good idea, but I don't know how to do that. I didn't see anything obvious in [the docs](https://cli.vuejs.org/guide/html-and-static-assets.html#the-index-file).

beekley: There doesn't seem to be a good way to dynamically set this title ([options](https://stackoverflow.com/questions/62023604/where-to-find-or-how-to-set-htmlwebpackplugin-options-title-in-project-created-w)). It only appears for a moment before router.ts sets it with a better value, so I think any more work here is over-optimization.

But I moved the rest of the titles out into a messages file.

client/src/App.vue
beekley: I'll have tabs open for dev and local and sandbox environments and this would make it easier to distinguish.

justinebreuch: SGTM

justinebreuch (approved):

PR #80, opened 2022-07-29
Scaffold lead prediction page with empty placeholders 

justinebreuch:
## Description

Addresses: Create basic scorecard component with placeholder data

Scaffolds more of the lead prediction page. Adds three factors for lead score to the page. 

Screen Shot 2022-07-29 at 2 33 57 PM

### New

3 most important factors for lead prediction are on page now. Placeholders for images and text. 

## Testing and Reviewing

See image.

PR #82, opened 2022-07-29
Move title strings to own file

beekley:
## Description

I forgot to push this commit for https://github.com/BlueConduit/open-data-platform/pull/79. This moves the strings out to its own file so it's easy to change.

### DIFF
--- a/client/src/assets/messages/common.ts
+++ b/client/src/assets/messages/common.ts
@@ -1,7 +1,11 @@
 // Add any common strings here.
 
-// Titles for the various pages in the app
-export const APP_TITLE = 'LeadOut';
-export const HOME_TITLE = 'Home';
-export const MAP_TITLE = 'Nationwide Lead Map';
-export const CONTACT_TITLE = 'Contact Us';
+/**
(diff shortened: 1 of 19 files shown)


client/src/assets/messages/common.ts
```diff
@@ -0,0 +1,7 @@
+// Add any common strings here.
+
+// Titles for the various pages in the app
```
justinebreuch, line 3: I put mine in a class of messages for organization. See prediction messages here: https://github.com/BlueConduit/open-data-platform/pull/80/files#diff-4c8fb56b2ab1f451991024f224e1a2cc85e2519b19dfa59ea18e97ebfeafa450

justinebreuch:
Thanks for doing this!

justinebreuch (approved):

kailamjeter (approved):

client/src/assets/messages/common.ts
beekley: That looks better so I've switched it to be the same.

PR #83, opened 2022-08-01
/watersystem/{pws_id} and /zipcode/scorecard/{zip_code}

justinebreuch:
## Description

Addresses: API skeleton (AWS api gateway + lambda creation). Specifically /watersystem/{pws_id} and /zipcode/scorecard/{zip_code}

### New

/get handlers for both water system and zipcode/scorecard

## Testing and Reviewing

AWS console gateway test events

### DIFF
--- a/api/src/watersystem/get.handler.ts
+++ b/api/src/watersystem/get.handler.ts
@@ -1,0 +1,89 @@
+import { APIGatewayProxyResult } from 'aws-lambda';
+import * as AWS from 'aws-sdk';
+import { RDSDataService } from 'aws-sdk';
+import {
+  ExecuteStatementRequest,
+  FieldList,
+  SqlParametersList,
+} from 'aws-sdk/clients/rdsdataservice';
+
+const CORS_HEADERS = {
+  'Access-Control-Allow-Headers': 'Content-Type',
+  'Access-Control-Allow-Origin': '*',
+  'Access-Control-Allow-Methods': 'OPTIONS,GET',
+};
+const SCHEMA = 'public';
+
+async function getWaterSystemData(
+  rdsService: RDSDataService,
+  params: SqlParametersList,
+): Promise<WaterSystemApiResponse> {
+  const executeParams: ExecuteStatementRequest = {
+    database: process.env.DATABASE_NAME ?? 'postgres',
+    resourceArn: process.env.RESOURCE_ARN ?? '',
+    schema: SCHEMA,
+    secretArn: process.env.CREDENTIALS_SECRET ?? '',
+    sql: `SELECT pws_id,
+                 lead_connections_count,
+                 service_connections_count
+          FROM water_systems
+          WHERE pws_id = :pws_id LIMIT 1`,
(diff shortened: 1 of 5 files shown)


api/src/watersystem/get.handler.ts
```diff
@@ -0,0 +1,89 @@
...
+      name: 'pws_id',
+      value: { stringValue: pwsId },
+    },
+  ];
+  let body: WaterSystemApiResponse;
+
+  try {
+    body = await getWaterSystemData(db, params);
+  } catch (error) {
+    console.log(`Error fetching water system data for pws id ${pwsId}`, error);
```
kailamjeter, line 63: should we return not 200 error in this case? not sure what best practice is

cdk/src/open-data-platform/data-plane/data-import/utils/lambda-function-factory.ts
```diff
@@ -35,3 +36,29 @@ export const lambdaFactory = (
...
   });
 };
+
+/**
+ * Produces a lambda function that requires db access to import data.
+ * @param scope: the construct under which to provision resource.
+ * @param props: schema identifiers passed to the function when writing data.
+ * @param id: ID to assign resource.
+ */
+export const apiLambdaFactory = (
```
kailamjeter, line 46: weirdly not seeing code highlighted here - do you know why that is? is this commented out?

kailamjeter (approved):

api/src/watersystem/get.handler.ts
```diff
@@ -0,0 +1,89 @@
...
+  const executeParams: ExecuteStatementRequest = {
+    database: process.env.DATABASE_NAME ?? 'postgres',
+    resourceArn: process.env.RESOURCE_ARN ?? '',
+    schema: SCHEMA,
+    secretArn: process.env.CREDENTIALS_SECRET ?? '',
+    sql: `SELECT pws_id,
+                 lead_connections_count,
+                 service_connections_count
+          FROM water_systems
+          WHERE pws_id = :pws_id LIMIT 1`,
```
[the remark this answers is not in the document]
justinebreuch: Good idea

cdk/src/open-data-platform/data-plane/data-import/utils/lambda-function-factory.ts
justinebreuch: Hmm no but the cdk deploy works so it should be okay

api/src/watersystem/get.handler.ts
justinebreuch: This would happen if there's an internal error or an exception through by RDS service, so we would want to return a 500. According to [this](https://aws.amazon.com/blogs/compute/error-handling-patterns-in-amazon-api-gateway-and-aws-lambda/) an error thrown should get passed through the response and then we can map certain responses to statuses. I'm adding a task to the backlog to see if more work needs to be done here to get errors handled elegantly. Thanks for raising!

PR #84, opened 2022-08-01
Move map into a view.

beekley:
## Description

It seems to be good practice [[1](https://vueschool.io/articles/vuejs-tutorials/how-to-structure-a-large-scale-vue-js-application/), [2](https://stackoverflow.com/a/50866150)] to have view components in their own directory, since they aren't used like other components.

This PR moves our (currently) only view into a `views` directory.

### New

- `views` directory of view components used by Vue-router.

### Changed

- some light component renaming.
- lint stuff. Prettier [deliberately doesn't respect](https://github.com/prettier/prettier/issues/5828#issuecomment-519371278) single quotes in the Vue HTML (e.g. template or script). I'm using [ESLint](https://eslint.vuejs.org/rules/html-quotes.html) + [this extension](https://marketplace.visualstudio.com/items?itemName=rohit-gohri.format-code-action) to run prettier then override it with ESLint. But I'd rather just use ESLint and throw away prettier.

## Testing and Reviewing

Manually poked around and it has no functional change.


client/src/components/NationwideMap.vue
```diff
@@ -1,9 +1,9 @@
 <template>
-  <div id='map-container'></div>
+  <div id="map-container"></div>
```
kailamjeter, line 2: My prettier has been making all strings single quote - what config are you using to get double quotes in html elements?

kailamjeter (approved):

client/src/components/NationwideMap.vue
justinebreuch: Can we keep single?

justinebreuch (approved):

client/src/components/NationwideMap.vue
beekley: Done with a hacky fix.

PR #85, opened 2022-08-03
Populate prediction panel with real data

justinebreuch:
## Description

Addresses: Implement API calls to use geo ids to fetch scorecard data 

Replaces static data in prediction panel with real prediction for the water system. Uses the first call to /geolocate to get pws_id and call /watersystem/{pwsId} to get prediction and show it 

## Testing and Reviewing
Screen Shot 2022-08-03 at 11 50 39 AM


client/src/api/api_client.ts
```diff
@@ -4,22 +4,46 @@ import axios from 'axios';
  * Client to interface with API.
  */
 class ApiClient {
-  static API_URL = 'https://ei2tz84crb.execute-api.us-east-2.amazonaws.com/dev/geolocate/';
+  static API_URL = 'https://ei2tz84crb.execute-api.us-east-2.amazonaws.com/dev';
   /**
    *   Retrieve geo ids of the geographic entities which intersect with the
    *   lat, long pair.
    */
   getGeoIdsFromLatLong = async (lat: string, long: string): Promise<ApiResponse> => {
```
kailamjeter, line 12: nit: I know this wasn't added here, but could we make it just getGeoIds (consistent with getWaterSystem and generally I think avoid putting param names in function names since they are self explanatory)

client/src/assets/messages/scorecard_summary_messages.ts
```diff
@@ -7,11 +7,13 @@ export class ScorecardSummaryMessages {
...
-    'The year your home was built influences the likelihood it was ' +
-    'constructed with lead pipes';
+    'The year your home was built influences the likelihood it was constructed with lead pipes';
   static INCOME_LEVEL = 'Income level';
-  static INCOME_LEVEL_EXPLAINED =
-    'Income levels correlate to the likelihood ' + 'of lead. Learn more.';
+  static INCOME_LEVEL_EXPLAINED = 'Income levels correlate to the likelihood of lead. Learn more.';
+  static LEADOUT = 'LeadOut by BlueConduit';
+  static LEAD_LIKELIHOOD_EXPLAINED =
+    'Based on age of homes, historical service line data, and information ' +
```
kailamjeter, line 15: seems like this string is broken off into a new line prematurely

client/src/components/PredictionPanel.vue
```diff
@@ -1,39 +1,88 @@
...
+      ScorecardSummaryMessages,
     };
   },
+  computed: {
+    // Predicted estimate of lead.
+    percentLead(): number | null {
+      const leadServiceLines = this.leadState?.data?.leadServiceLines;
+      const serviceLines = this.leadState?.data?.serviceLines;
+
+      // Protect against dividing by 0
```
kailamjeter, line 56: super nit (sorry this is extremely googley) end comment with a period

client/src/components/PredictionPanel.vue
```diff
@@ -1,39 +1,88 @@
...
+      if (leadServiceLines != null && serviceLines != null && serviceLines != 0) {
+        return Math.round(leadServiceLines / serviceLines);
+      }
+      return null;
```
kailamjeter, line 60: do we want to show a different message if this is null?

client/src/components/PredictionPanel.vue
```diff
@@ -1,39 +1,88 @@
...
+    },
+    pwsId(): string | null {
+      return this.geoState?.geoids?.pwsId ?? null;
```
kailamjeter, line 63: same here - should we indicate that there is no pwsId in the message? fine to leave a TODO for both

client/src/components/PredictionPanel.vue
```diff
@@ -1,39 +1,88 @@
...
+    },
+  },
+  watch: {
+    // Listen for changes to pws id. Once it changes, a new prediction
+    // must be fetched.
+    geoState: function() {
```
kailamjeter, line 69: does this only get triggered when pwsId changes? seems like it could get triggered if another geoid changes in which case we wouldn't want to refetch ws prediction... 

if not, maybe try watching on geoState?.geoids?.pwsId with a deep watcher?

client/src/model/slices/lead_data_slice.ts
```diff
@@ -0,0 +1,51 @@
...
+const client = new ApiClient();
+
+// Automatically generates action creators and action types that correspond
+// to the reducers and state. See: https://redux-toolkit.js.org/api/createslice
+// TODO(breuch): Consider showing user message on error.
+const leadDataSlice = createSlice({
+  name: 'leadDataSlice',
+  initialState,
+  reducers: {
+    waterSystemQueried(state: LeadDataState, action: PayloadAction<LeadData>) {
```
kailamjeter, line 17: will these contain separate logic in the future? seems like they could be combined to a updateStateWithLeadData action instead?

client/src/model/store.ts
```diff
@@ -1,10 +1,29 @@
...
+const store = configureStore({
+  reducer: rootReducer,
+});
 
-const store = configureStore({ reducer: { geosReducer } });
 // Emits events for reducers.
 export const dispatch = store.dispatch;
 
+// Same as https://react-redux.js.org/api/hooks#useselector
+export const useSelector = <State extends RootState = RootState>(
```
kailamjeter, line 21: Can you add to this comment 'Allows you to extract data from the Redux store state.'

client/src/components/PredictionPanel.vue
```diff
@@ -1,39 +1,88 @@
...
   name: 'PredictionPanel',
   components: { MapGeocoderWrapper },
+  setup() {
+    // Listen to state updates.
+    const geoState = useSelector((state) => state.geos) as GeoState;
+    const leadDataState = useSelector((state) => state.leadData) as LeadDataState;
+
+    return {
+      geoState: geoState,
+      leadState: leadDataState,
```
kailamjeter, line 40: out of curiosity - so generally this is how we want to listen to store changes? and no components will look at store directly, this is hidden in useSelector?

kailamjeter (approved):

client/src/api/api_client.ts
justinebreuch: Yeah good point thanks!

client/src/components/PredictionPanel.vue
justinebreuch: All good - don't feel bad about nits

client/src/assets/messages/scorecard_summary_messages.ts
justinebreuch: It only seemed that way because HOME_AGE_EXPLAINED was not properly pruned

client/src/components/PredictionPanel.vue
justinebreuch: This seems to be a suggested way of doing it to avoid duplicated code but this "as XState" type casting is not ideal. The issue is we don't have the react hooks that give these states back typed

client/src/components/PredictionPanel.vue
justinebreuch: Null state signifies that a search hasn't been made so I need to add handling for API errors. So when predictions don't get returned. Added a todo and this is already in the eng task list

client/src/components/PredictionPanel.vue
justinebreuch: Same here

client/src/components/PredictionPanel.vue
justinebreuch: Totally agree. So when I tried watching on anything deeper than geostate it wasn't working even with deep. I also tried doing a computed property and watching on that but it never got hit

client/src/model/slices/lead_data_slice.ts
justinebreuch: We can add action types to the actions and do one reducer with a switch statement. Let me look into this and do a fast-follow

client/src/model/store.ts
justinebreuch: Yep!

PR #86, opened 2022-08-03
Update state with lat,long in URL

kailamjeter:
Pulls lat,long from the URL if it exists and dispatches event to query API for scorecard information. 
Updates map router path to expect the latlong string.


client/src/App.vue
```diff
@@ -48,6 +48,16 @@ export default defineComponent({
   },
   watch: {
     $route(to: RouteLocation) {
```
justinebreuch, line 50: Can you link to how this gets wired up like how route get here? Good chance I'm missing something too

kailamjeter: Done!

PR #87, opened 2022-08-04
Initial landing page

beekley:
## Description

This PR adds some relatively-simple, mostly-unstyled components that implement [these mocks](https://www.figma.com/proto/f6DyufrwAASbLVK2dLmgPe/Homepage-Options?page-id=0%3A1&node-id=2%3A1256&viewport=407%2C352%2C0.12&scaling=min-zoom&starting-point-node-id=2%3A1181&show-proto-sidebar=1).

I don't really like how the CSS is written and would rather switch to a component library (TBD). I'm also not super into the `LandingPageSection` component, which generalizes the landing page stuff.

### New

- Landing page with:
  - Search bar that routes to the map for the searched address.
  - Some other components for displaying info.

### Changed

- Some small changes to make the UI match the mocks more (e.g. "contact" -> "about").

## Testing and Reviewing

Deployed locally and I can follow each of the links.


client/src/components/ScorecardSearch.vue
```diff
@@ -0,0 +1,45 @@
...
+      long: 0,
+    };
+  },
+  methods: {
+    onGeocodeResults(lat: number, long: number) {
+      this.lat = lat;
+      this.long = long;
+    },
+    onSearch() {
+      window.location.replace(`${MAP_ROUTE_BASE}/${this.lat},${this.long}`);
```
justinebreuch, line 39: Could you use the router here or is the recommended way of doing it?

beekley: Good idea, done (in a future commit).

client/src/assets/messages/landing.ts
```diff
@@ -0,0 +1,30 @@
+// Strings for the landing page.
+
+/**
+ * Strings related to the scorecard-related part of the landing page.
+ */
+export class ScorecardSearch {
+  static HEADER = `Get your drinking water score`;
+  static BODY = `Understand the likelihood of your home or community's exposure to lead service lines.`;
+  static CTA_PLACEHOLDER = `Enter your address`;
+  static CTA_BUTTON = 'Get Score';
```
justinebreuch, line 10: This is super nit but either use `` or '' for consistency. Here and below

client/src/assets/messages/landing.ts
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Strings related to the filter-related part of the landing page.
+ */
+export class FilterInfo {
+  static SUPER_HEADER = `Take Action`;
+  static HEADER = `Chosing a water filter`;
+  static BODY = `There are several types of water filters. Choosing the right one depends on your budget and lifestyle.`;
+  static CTA_BUTTON = 'Learn more about water filters';
+}
```
justinebreuch, line 30: If there's multiple exports, I've been doing export {FilterInfo, MapInfo, etc..} at the bottom bc it's been helping me to know what's available from every fail at a quick glance

client/src/components/LandingPageFooter.vue
```diff
@@ -0,0 +1,60 @@
+<template>
+  <div class='landing-page-footer'>
+    <div class='left-align'>
+      <div v-for='route in routes' :key='route[0]'>
+        <router-link :to='route[1]'>{{ route[0] }}</router-link>
+      </div>
+    </div>
+    <!-- Replace with Logo -->
```
justinebreuch, line 8: Logo's in assets/ folder!

client/src/components/LandingPageFooter.vue
```diff
@@ -0,0 +1,60 @@
...
+  </div>
+</template>
+
+<script lang="ts">
+import { defineComponent } from 'vue';
+import { Titles } from '../assets/messages/common';
+import * as router from '../router';
+
+/**
+ * A component for searching for a scorecard.
```
justinebreuch, line 19: Nit:  Replace comment for this component

client/src/components/LandingPageSection.vue
```diff
@@ -0,0 +1,47 @@
...
+  CTA_PLACEHOLDER?: string;
+  CTA_BUTTON: string;
+}
+
+/**
+ * A component that displays basic content on the landing page. It can accept inner components for
+ * extra functionality, such as buttons, links, or videos. The purpose of this component is to
+ * keep the landing page DRY and define style in one place.
+ */
+export default defineComponent({
```
justinebreuch, line 27: Same with exports here

client/src/components/LandingPageSection.vue
```diff
@@ -0,0 +1,47 @@
...
+  name: 'LandingPageSection',
+  // TODO: add props to customize basic styling. E.g. background color.
+  props: {
+    messages: {
+      type: Object as () => Messages,
```
justinebreuch, line 32: Does Object as PropType<Messages> work?

client/src/components/ScorecardSearch.vue
```diff
@@ -0,0 +1,45 @@
...
+import { defineComponent } from 'vue';
+import { MAP_ROUTE_BASE } from '../router';
+import { Messages } from './LandingPageSection.vue';
+import GeocoderInput from './GeocoderInput.vue';
+
+/**
+ * A component for searching for a scorecard.
+ */
+export default defineComponent({
+  name: 'ScorecardSearch',
```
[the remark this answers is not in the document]
beekley: I did and actually think it's fine to have several wrappers which each have their own purpose. This one opens the map, while MapGeocoderWrapper calls an API which are different enough that I don't think they'd make sense to combine.

The common code in GeocoderInput keeps it DRY enough.

client/src/assets/messages/landing.ts
beekley: I'm don't love that convention (mainly since I'd rather see if something is exported in-line with its definition). The [TS style guide](https://g3doc.corp.google.com/javascript/typescript/g3doc/dev/readability/styleguide.md?cl=head#exports) doesn't seem to have an opinion and CS has plenty of both ([1](https://source.corp.google.com/search?q=lang:ts%20%22export%20%22&sq=), [2](https://source.corp.google.com/search?q=lang:ts%20%22export%20%7B%22&start=11)).

Maybe let do it case-by-case, depending on the size+complexity of a file? This is small enough and exports everything, so it's still easy to see exports at a glance.

client/src/components/LandingPageSection.vue
beekley: See above.

client/src/assets/messages/landing.ts
beekley: Done.

client/src/components/LandingPageSection.vue
justinebreuch: Do you prefer this syntax? We have the PropType syntax elsewhere taken from https://vuejs.org/guide/typescript/options-api.html#typing-component-props

beekley: oh, I didn't realize you were suggesting another approach. I like that one better so I'll switched to using PropType.

kailamjeter (approved):

PR #90, opened 2022-08-04
Replace static data in the summary panel with real data 

justinebreuch:
## Description

Addresses: Implement API calls to get demographic comparisons for zip codes

Replaces static comparison data with real API response for /zipcode/scorecard

I need to update the API to return the US aggregated table and then change these columns accordingly

## Testing and Reviewing
In UI


client/src/components/ScorecardSummaryPanel.vue
```diff
@@ -40,12 +42,47 @@ export default defineComponent({
   components: {
     ScorecardSummaryRow,
   },
+  setup() {
+    // Listen to state updates.
+    const geoState = useSelector((state) => state.geos) as GeoState;
+    const demographicData = useSelector((state) => state.demographicData) as DemographicDataState;
```
kailamjeter, line 48: nit/optional: the name of this feels a bit inconsistent with the geoState because of the Data portion. The state holds demographic data, same as geo state holds geographic data. It might be more appropriate to use 'data' in all state names or none.

client/src/model/states/model/demographic_data.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Information about demographics at geo-granularity.
+ */
+export interface DemographicData {
+  geoId?: string;
```
kailamjeter, line 5: Can you comment as to what this might be? i.e. explain this is census geo id of a zip, county, state

kailamjeter (approved):

client/src/model/states/model/demographic_data.ts
justinebreuch: Done!

client/src/components/ScorecardSummaryPanel.vue
justinebreuch: yeah good point. Renamed

PR #91, opened 2022-08-05
Add /parcel/{lat,long} endpoint to API that returns predictions for parcels we have

justinebreuch:
## Description

Addresses: Add /parcel/{lat,long} endpoint to API that returns predictions for parcels we have

### New
Lambda handler for parcel data
Gateway endpoint for parcel lambda

## Testing and Reviewing

API gateway test event


beekley:
(I had asked this is the design doc too) Why does this use `lat,long` as the input instead of an ID like ZIP ([`zip_code`](https://github.com/BlueConduit/open-data-platform/blob/main/api/src/zipcode/scorecard/get.handler.ts#L47)) or Water Systems ([`pws_id`](https://github.com/BlueConduit/open-data-platform/blob/main/api/src/watersystem/get.handler.ts#L46)) do? Using an ID would make parcels more like the other geospatial entities.

justinebreuch:
> (I had asked this is the design doc too) Why does this use `lat,long` as the input instead of an ID like ZIP ([`zip_code`](https://github.com/BlueConduit/open-data-platform/blob/main/api/src/zipcode/scorecard/get.handler.ts#L47)) or Water Systems ([`pws_id`](https://github.com/BlueConduit/open-data-platform/blob/main/api/src/watersystem/get.handler.ts#L46)) do? Using an ID would make parcels more like the other geospatial entities.

Added a todo about standardized addresses as a key. For context, the addresses in the DB don't match the geo coder output so it would be challenging to index into this table by address. I'm sure there is a solution here if that's preferred but will require some further implementation / design

api/src/parcel/get.handler.ts
```diff
@@ -0,0 +1,96 @@
...
+import { ExecuteStatementRequest, SqlParametersList } from 'aws-sdk/clients/rdsdataservice';
+
+const CORS_HEADERS = {
+  'Access-Control-Allow-Headers': 'Content-Type',
+  'Access-Control-Allow-Origin': '*',
+  'Access-Control-Allow-Methods': 'OPTIONS,GET',
+};
+const SCHEMA = 'public';
+const SQL_QUERY = `SELECT address,
+                          public_lead_prediction,
```
kailamjeter, line 13: should these be low/high estimates?

api/src/parcel/get.handler.ts
```diff
@@ -0,0 +1,96 @@
...
+    schema: SCHEMA,
+    secretArn: process.env.CREDENTIALS_SECRET ?? '',
+    sql: SQL_QUERY,
+    parameters: params,
+  };
+  let body: ParcelApiResponse = {};
+
+  const results = await rdsService.executeStatement(executeParams).promise();
+  for (let record of results.records ?? []) {
+    body = {
```
kailamjeter, line 36: small error handling thing, but should we check that record has all values before indexing? like if record.length < 3 print an error? I can imagine an index out of bounds error here requiring a bit of digging.

kailamjeter (approved):

api/src/parcel/get.handler.ts
justinebreuch: Done!

api/src/parcel/get.handler.ts
justinebreuch: Added the prediction lambda to this PR to read in the schema values

PR #94, opened 2022-08-10
Surface parcel predictions in the FE and fix routing issues 

justinebreuch:
## Description

Addresses: Implement API calls to use geo ids to fetch parcel predictions

Add getParcel to api client. PredictionPanel calls getParcel and displays the parcel prediction if an address was searched instead of a zip code 

### New

GeoType which, for now, identifies the granularity for a query
Added getParcel() to the client 

### Changed

Routing so that we have a home page, scorecard page, and a map page

## Testing and Reviewing

In UI


client/src/components/PredictionPanel.vue
```diff
@@ -47,18 +56,26 @@ export default defineComponent({
       ScorecardSummaryMessages,
     };
   },
+  // TODO: Handle error state where there is no lead prediction
+  // after the API has returned.
   computed: {
-    // Predicted estimate of lead.
-    percentLead(): number | null {
+    // Predicted lead likelihood for parcel's public lines.
+    publicLeadLikelihood(): string | null {
```
justinebreuch, line 63: Note that low prediction is the same as high prediction in the DB right now bc we don't yet have intervals

client/src/api/api_client.ts
```diff
@@ -8,6 +9,16 @@ class ApiClient {
   static API_URL = 'https://ei2tz84crb.execute-api.us-east-2.amazonaws.com/dev';
 
   request = async (endpoint: string, callback: (data: any) => any): Promise<ApiResponse> => {
+    axiosRetry(axios, {
+      retries: 3,
+      retryDelay: (retryCount) => {
+        return retryCount * 2000; // Time between retries
+      },
+      retryCondition: (error: AxiosError) => {
+        return error.response?.status === 502;
```
kailamjeter, line 18: can you add a comment as to why we retry on this error code?

client/src/components/PredictionPanel.vue
```diff
@@ -1,16 +1,28 @@
...
-      {{ percentLead }}% lead.
-    </div>
-    <div class='explain-text'>
-      {{ ScorecardSummaryMessages.LEAD_LIKELIHOOD_EXPLAINED }}
+    <div class='container-column center-container'>
+      <div class='explain-text'>
+        {{ ScorecardSummaryMessages.LEADOUT }}
+      </div>
+      <div class='h1-header semi-bold'
+           v-if='showWaterSystemPrediction'>
```
kailamjeter, line 11: silly q - where is parcel message?

client/src/components/PredictionPanel.vue
```diff
@@ -47,46 +60,83 @@ export default defineComponent({
...
+    showWaterSystemPrediction(): boolean {
+      return this.geoState?.geoids?.geoType != GeoType.address && this.pwsId != null && this.percentLead != null;
+    },
   },
   watch: {
     // Listen for changes to pws id. Once it changes, a new prediction
     // must be fetched.
-    'geoState.geoids.pwsId': function() {
-      if (this.geoState?.geoids?.pwsId != null) {
+    'geoState.geoids': function() {
```
kailamjeter, line 101: was this the fix? i.e. listen on all geo IDs and not just pws id

client/src/model/states/model/geo_data.ts
```diff
@@ -1,9 +1,24 @@
+/**
+ * Represent the granularity of a geo query.
+ *
+ * Must match mapbox.
+ * TODO: Unify with geographic level somehow.
+ */
+enum GeoType {
+  address = 'address',
+  postcode = 'postcode',
```
kailamjeter, line 9: is this zip? if so can we call it zipcode

kailamjeter (approved):

client/src/api/api_client.ts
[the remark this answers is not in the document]
justinebreuch: Done! 502 is an internal error

client/src/components/PredictionPanel.vue
justinebreuch: Added a variable "showParcelPrediction" for clarity

client/src/model/states/model/geo_data.ts
justinebreuch: This is annoying but it has to match the mapbox place types. I can't find the documentation for it but the result that geocoder returns has a field called "place_type" and when it's zipcode it's called a postcode and when it's address it's address. I need to check what county / city / state produces

client/src/components/PredictionPanel.vue
justinebreuch: I updated it to just geos so I could dispatch events on both pws id and geotype/lat/long

PR #95, opened 2022-08-10
Add S3 triggers to launch data import lambdas when files are added

justinebreuch:
An imperfect solution to get some data in the dev db.

Anytime there is an update to our s3 bucket, the data import functions will all trigger. This is imperfect because some of them need to be run multiple times with event properties passed in. 

Will keep looking into this, but want to unblock our 20% engineer, so he can see some data and the dev environment can be shared with the rest of the team. 

These triggers are in cdk/src/open-data-platform/data-plane/data-import/data-import-stack.ts

Also re-factored violations data importer so use a common handler factory, which is less important but I was touching it anyway.


justinebreuch:
Looping in Aaron, the AWS sensei, in case he has ideas

cdk/src/open-data-platform/data-plane/data-import/handler-factory.ts
```diff
@@ -52,12 +52,17 @@ export const geoJsonHandlerFactory =
...
+      rowOffset,
+      batchSize,
+      s3Params,
+    });
 
     const db = new AWS.RDSDataService();
     let results: ProcessResult = {
       processedBatchCount: 0,
-      sucessfulBatchCount: 0,
+      successfulBatchCount: 0,
```
beekley, line 65: Just an idea: Whenever a lambda invocation finishes, it could invoke another instance with a larger `rowOffset` until `successfulBatchCount == 0`.

beekley (approved):

cdk/src/open-data-platform/data-plane/data-import/data-import-stack.ts
```diff
@@ -67,6 +71,12 @@ export class DataImportStack extends Construct {
         let role = iam.Role.fromRoleArn(scope, id + '-' + f, f.role.roleArn);
         cluster.grantDataApiAccess(role);
       }
+
+      // Trigger all the data import lambdas when files change.
+      const destination = new notification.LambdaDestination(f);
```
kailamjeter, line 76: I wonder if we can also match on filename here to update on like an OBJECT_MODIFIED event - like if f.s3params.key == modified filename?

Not urgent just a thought, fine to leave a TODO

kailamjeter (approved):
Thanks for doing this!!

cdk/src/open-data-platform/data-plane/data-import/data-import-stack.ts
justinebreuch: It seems like they only have created and deleted events which is unfortunate. So we need to make sure we delete it temporarily and then re-upload https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-how-to-event-types-and-destinations.html

justinebreuch: I updated this to use a mapping between the ids for the lambda functions : the s3 files. There's some remaining TODOs but now the functions trigger only on relevant files.

cdk/src/open-data-platform/data-plane/data-import/handler-factory.ts
```diff
@@ -49,23 +49,46 @@ export const geoJsonHandlerFactory =
...
       erroredBatchCount: 0,
     };
 
-    try {
-      results = await readGeoJsonFile(db, s3Params, callback, rowOffset, rowLimit, batchSize);
-    } catch (error) {
-      console.log(`Error after processing ${results.processedBatchCount} batches:`, error);
-      throw error;
+    let keepReading = true;
+    while (keepReading) {
```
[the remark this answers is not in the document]
justinebreuch: I added a description to the "Set up automatic DB updates" task as follows: Ideal workflow is 1 s3 trigger on all files that launch a single lambda that splits up the file into reasonably sized chunks and then writes those files to a new S3 bucket. Then triggers on those new S3 files to run individual lambda functions. 

This would mean we should not have to invoke the lambdas multiple times bc files should be small enough to be parsable. I also removed my white loop because some of the numbers I was getting from results were not making sense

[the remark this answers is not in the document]
beekley: SGTM.

beekley (approved):

cdk/src/open-data-platform/data-plane/data-import/write-water-systems-data.handler.ts
```diff
@@ -58,29 +58,49 @@ function getValueOrDefault(field: string): number {
...
-      .serviceConnectionsCount(getValueOrDefault(properties.service_connections_count))
-      .populationServed(getValueOrDefault(properties.population_served_count))
-      // Keep JSON formatting. Post-GIS helpers depend on this.
-      .geom(JSON.stringify(value.geometry))
-      .build()
-  );
+
+  console.log(`properties: ${JSON.stringify(properties)}`);
+
+  if (value.geometry != null) {
```
kailamjeter, line 67: nit: prefer to check for null and return null first (just easier to see whats going on in the null case)

cdk/src/open-data-platform/data-plane/data-import/write-water-systems-data.handler.ts
```diff
@@ -58,29 +58,49 @@ function getValueOrDefault(field: string): number {
...
+    const reportedLeadConnections = getValueOrDefault(properties.lead_connections);
+    let leadConnectionsCount;
+
+    // If there is a reported lead count, use that
+    if (reportedLeadConnections > 0) {
+      leadConnectionsCount = reportedLeadConnections;
+      // Otherwise, get estimate from prediction interval.
+    } else {
+      leadConnectionsCount =
+        (getValueOrDefault(properties.low) + getValueOrDefault(properties.high)) / 2;
```
kailamjeter, line 77: why store an avg instead of low/high?

cdk/src/open-data-platform/data-plane/data-import/write-water-systems-data.handler.ts
justinebreuch: Done!

cdk/src/open-data-platform/data-plane/data-import/write-water-systems-data.handler.ts
justinebreuch: Added a TODO but basically storing these values would require changing a bunch of tables so we need to do this separately

PR #96, opened 2022-08-11
Serve APIs with CloudFront

beekley:
## Description

Currently, the API lambdas are publicly accessible through API Gateway. However, API Gateway provides it's own hostname, which is [hardcoded into the client](https://github.com/BlueConduit/open-data-platform/blob/f6e8ca0f39f2887ff13496dc1985702f3e3b7ef9/client/src/api/api_client.ts#L9). Since this isn't the same hostname as the client, we'd have to dynamically tell the client where the API gateway is. (plus we'd have to have broader CORS and other auth control).

A solution to this is to serve the API through CloudFront (CF). This PR configures CloudFront to direct requests to `/api/v1/...` to the API Gateway (and then to the API lambdas). So the client can make API calls without changing hostnames.

This previously set up each lambdas as its own CF origin, but that was ugly since it forced lamdas to do stuff they weren't built for (e.g. URL path param parsing).

### New

- CF behaviors that point to each of the lambdas.
- Some descriptive error messages in the APIs.

### Changed

- The client makes requests to the API behind CF.
- Moved the API stack to the app plane, since it's part of the user-facing application layer (like the tile server).
- All URL path prefixes (e.g. `api` or `tiles/v1`) are stored in a single location, even for the client.

## Testing and Reviewing

The client is successfully making these requests to the API, which is returning values. I manually spot checked individual APIs for data I know is in the DB.

* Scorecard: [Page](https://beekley.leadout-sandbox.blueconduit.com/scorecard/40.590405,-74.415312?layer=lead-service-lines-by-water-system-style).
* Geolocate: [Results](https://beekley.leadout-sandbox.blueconduit.com/api/geolocate/41.256231,-96.012877).
* Water System: [Results](https://beekley.leadout-sandbox.blueconduit.com/api/watersystem/WA5333675).
* Parcel: [Results](https://beekley.leadout-sandbox.blueconduit.com/api/parcel/41.72417,-83.556453).
* ZIP: [Results](https://beekley.leadout-sandbox.blueconduit.com/api/zipcode/scorecard/01039).


cdk/src/open-data-platform/app-plane/api/lambda-function-factory.ts
```diff
@@ -0,0 +1,61 @@
...
+      nodeModules: ['moment'],
+    },
+    role: props.role,
+  });
+  props.credentialsSecret.grantRead(f);
+  return {
+    lambda: f,
+    // Assumes that there is some URL parameter after the path.
+    path: `/${id}/*`,
+    // This might be better handled in the frontend stack, which can lock down auth and CORS.
```
kailamjeter, line 53: can we make this a TODO or remove?

kailamjeter (approved):

api/src/util.ts
```diff
@@ -0,0 +1,9 @@
+import prefixes from '../../cdk/src/open-data-platform/frontend/url-prefixes';
+
+export const CORS_HEADERS = {
+  'Access-Control-Allow-Headers': 'Content-Type',
+  'Access-Control-Allow-Origin': '*',
```
arockett, line 5: Seems like a very permissive CORS origin config, any reason not to dynamically set CORS based on the apps domain?

api/src/parcel/get.handler.ts
```diff
@@ -51,13 +47,20 @@ async function getParcelData(
...
-  const coordinates = event.pathParameters?.latlong ?? '';
+export const handler = async (event: APIGatewayProxyEventV2): Promise<APIGatewayProxyResult> => {
+  // Parse out the path parameters.
+  const coordinates = trimPath(event.rawPath)[2];
+  if (!coordinates || coordinates == '')
+    return {
+      statusCode: 400,
+      headers: CORS_HEADERS,
+      body: 'No coordinates provided. Expected {lat},{long}.',
+    };
```
arockett, line 58: Did API Gateway previously handle path validation? Is that why we need to parse and validate in code now?

cdk/src/open-data-platform/app-plane/api/lambda-function-factory.ts
```diff
@@ -0,0 +1,61 @@
...
+    vpc: props.vpc,
+    vpcSubnets: { subnetType: ec2.SubnetType.PRIVATE_WITH_NAT },
+    environment: {
+      CREDENTIALS_SECRET: props.credentialsSecret.secretArn,
+      DATABASE_NAME: props.db,
+      RESOURCE_ARN: props.cluster.clusterArn,
+    },
+    bundling: {
+      externalModules: ['aws-sdk'],
+      nodeModules: ['moment'],
```
arockett, line 44: For another time but just fyi: moment is EOL and the project owners have a page with recommended alternatives. We've chosen to go with 'luxon' moving forward. Can you add a TODO comment here for switching to luxon so we have code context for this change?

cdk/src/open-data-platform/app-plane/api/lambda-function-factory.ts
```diff
@@ -0,0 +1,61 @@
...
+    lambda: f,
+    // Assumes that there is some URL parameter after the path.
+    path: `/${id}/*`,
+    // This might be better handled in the frontend stack, which can lock down auth and CORS.
+    url: f.addFunctionUrl({
+      // TODO: consider limiting this to only allow the CloudFront role to make this request.
+      authType: FunctionUrlAuthType.NONE,
+      // TODO: limit to blueconduit.com.
+      cors: { allowedOrigins: Cors.ALL_ORIGINS },
+    }).url,
```
arockett, line 59: Is there a plan to publish this API for external use either publicly with rate limiting or via API keys? Seems like more work to manage CORS and auth manually for each function as opposed to centrally via API Gateway.

cdk/src/open-data-platform/app-plane/api/lambda-function-factory.ts
```diff
@@ -0,0 +1,61 @@
...
+      externalModules: ['aws-sdk'],
+      nodeModules: ['moment'],
+    },
+    role: props.role,
+  });
+  props.credentialsSecret.grantRead(f);
+  return {
+    lambda: f,
+    // Assumes that there is some URL parameter after the path.
+    path: `/${id}/*`,
```
justinebreuch, line 52: How does this work when they want to add on endpoints? Like if you want to add /zipcode/{id} on top of /zipcode/scorecard/{id}

arockett (commented):
I don't have an OpenDataPlatform env setup to test this change, but the refactor to proxy API requests through CF and remove API Gateway looks like it will work. I'd like to better understand the medium-term requirements regarding the API, especially around auth and public access. I'm a little wary of exposing lambdas directly to the internet without restrictions.

justinebreuch (approved):

cdk/src/open-data-platform/app-plane/api/lambda-function-factory.ts
justinebreuch: The plan from the API design was to transition this into the developer API eventually, but I didn't know how to dynamically tell the client where the API gateway is, which I think is what lead to refactoring to cloudfront. Just wanted to note on the PR description, API Gateway did provide caching - I turned it on in my console when I was developing, but I never set it up in code. I'm assuming Cloudfront does this for free without adding a property?

api/src/parcel/get.handler.ts
beekley: Yes, and this is now obsolete.

cdk/src/open-data-platform/app-plane/api/lambda-function-factory.ts
beekley: Done.

cdk/src/open-data-platform/app-plane/api/lambda-function-factory.ts
beekley: Obsolete.

cdk/src/open-data-platform/app-plane/api/lambda-function-factory.ts
beekley: Obsolete.

api/src/util.ts
beekley: I added a TODO to lock this down or remove the CORS headers altogether, since that might be handled at the gateway layer instead.

cdk/src/open-data-platform/app-plane/api/lambda-function-factory.ts
beekley: The root of this comment is obsolete since I added the API Gateway back. But to answer the literal question: yes and I added a task to set up a rate limit once we publish the API docs.

arockett (approved):
Looks good!

PR #97, opened 2022-08-11
Resource page

kailamjeter:
## Description

Adds resources page. 

Mock: https://www.figma.com/file/0LGxjZnPD3N0dCqHv8LRr6/NW-Map-Design-System-WIP?node-id=486%3A6401

Note that this does not add a link to this to the scorecard page. I will do this in a separate PR.

### New

- Resource view and components which contain content for the page.

## Testing and Reviewing

npm run serve

## screenshot

Screen Shot 2022-08-11 at 3 45 59 PM


client/src/components/ImageCard.vue
```diff
@@ -0,0 +1,74 @@
+<template>
+  <div class='card'>
+    <span class='asset'>
+      <img :src='require(`@/assets/media/${content.image ?? defaultImage}`)'
+           alt=''>
+    </span>
+    <div class='header-1'>{{ content.header }}</div>
```
beekley, line 7: nit: It's more semantic to use `h1` and `h2` instead of `div`s with classs names here.

beekley (approved):
From making the landing page and looking at this, I wonder if we'd be better served using some off-the-shelf static site generator to build these pages (e.g. [this vue-based one](https://vue-community.org/guide/ecosystem/static-site-generators.html#vuepress)). That way we wouldn't have to write components to display static text like this. And instead we'd just write the content as markdown.

client/src/components/ImageCard.vue
kailamjeter: good point - I initially had these then realized they were messing up my styling, but just added more styling to normalize. Thanks!

PR #99, opened 2022-08-12
Add action sections to scorecard page linking to resources and national map pages

justinebreuch:
## Description

Addresses: Add share / additional resources section to scorecard page and Add Actions to Take section

### New

Adds ActionSection component that we need for the scorecard and error page. Includes a header, subheader and CTA button. 

### Changed

Moved footer out of landing page into all pages. The map page needs to be made larger to fit the screen in a separate PR. 

## Testing and Reviewing

See screenshots
Screen Shot 2022-08-12 at 12 44 37 PM
Screen Shot 2022-08-12 at 12 44 46 PM


kailamjeter:
Does this cover the explanation of the score section? looks like it in the screenshot

client/src/components/ActionSection.vue
```diff
@@ -0,0 +1,67 @@
...
+  gap: 20px;
+  height: var(--height);
+  padding: 20px;
+}
+
+.outline-button {
+  color: var(--button-color)
+}
+
+
```
kailamjeter, line 66: nit: remove whitespace here

client/src/components/PageFooter.vue
```diff
@@ -39,6 +39,7 @@ export default defineComponent({
 
 <style scoped>
 .landing-page-footer {
```
kailamjeter, line 41: can you remove 'landing' from the classnames here too

kailamjeter (approved):

client/src/components/ActionSection.vue
justinebreuch: Done!

client/src/components/PageFooter.vue
justinebreuch: Nice catch

justinebreuch:
> screenshot

I just refreshed the text based on mocks but that section was there before

PR #102, opened 2022-08-13
Make geolocate/ return bounding box with geoid

RP-3:
## Description

Addresses: [story](https://tables.area120.google.com/table/bX_kMNBW8LA35n6l5Hz_co/row/9SjAy8XtQi30i0Vy55xOWx)

- Update client to fit map to provided bounding box by default (if available)
- Make geolocate/ return bounding box with geoid
- Prevent geolocate/ from swallowing failed db-query errors

### New

- N/A

### Changed

- Geolocate/ endpoint now returns a richer object including a bounding box instead of a simple string
- If multiple polygons match the provided coords, we sort by ID and return the lowest one. Previously was nondeterministic.
- The client also expects this bounding box to be passed around

### Removed

- N/A

## Testing and Reviewing

- I've only tested the one usecase I understand, which is:
  - Type in an address in the home page
  - Hit 'enter'
  - The map should zoom into the bounding box for the zipcode. The entire zipcode should fit on the map view.
- Note: The zooming (and general interactivity of the map) is pretty slow. I wonder if this is because the polygons we're rendering on it are way too detailed.

### DIFF
--- a/api/src/geolocate/get.handler.ts
+++ b/api/src/geolocate/get.handler.ts
@@ -28,6 +28,20 @@ async function getGeoDataForLatLong(
     resourceArn: process.env.RESOURCE_ARN ?? '',
     schema: SCHEMA,
     secretArn: process.env.CREDENTIALS_SECRET ?? '',
+    /*
+    Subquery: selects the one row with the lowest id with that also intersects
+      with the provided :long and :lat. Will return a bounding box in the form:
+      'NY0700789	'BOX(-74.258843 40.476578,-73.700169 40.917705)'. See ST_EXTENT
+      below.
+    Outer Query:
+      ST_EXTENT: Aggregates the geometries of all rows returned in the subquery
+      to return their combined bounding box. See:
+      https://postgis.net/docs/manual-1.5/ST_Extent.html
+      STRING_AGG: Since ST_EXTENT is an aggregate function (like avg(), max()),
+      all other selected rows also need to be aggregated. STRING_AGG just
+      concats all the returned rows together. However, this has no effect
+      since the subquery already ensures that only one row is returned.
+    */
     sql: `
       WITH target_row AS (
         SELECT ${geoid} AS id, geom AS geom
@@ -42,7 +56,6 @@ async function getGeoDataForLatLong(
       FROM target_row`,
     parameters: params,
   };
-  //Example response: NY0700789	'BOX(-74.258843 40.476578,-73.700169 40.917705)'
   const results = await rdsService.executeStatement(executeParams).promise();
   const geoID = results.records ? (results.records[0] || undefined) : undefined;
   if(!geoID) return undefined;
--- a/api/src/geolocate/get.handler.ts
+++ b/api/src/geolocate/get.handler.ts
@@ -3,7 +3,6 @@ import * as AWS from 'aws-sdk';
 import { RDSDataService } from 'aws-sdk';
 import {
   ExecuteStatementRequest,
-  FieldList,
   SqlParametersList,
 } from 'aws-sdk/clients/rdsdataservice';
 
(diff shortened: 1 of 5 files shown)


api/src/geolocate/get.handler.ts
```diff
@@ -21,24 +20,48 @@ async function getGeoDataForLatLong(
...
     database: process.env.DATABASE_NAME ?? 'postgres',
     resourceArn: process.env.RESOURCE_ARN ?? '',
     schema: SCHEMA,
     secretArn: process.env.CREDENTIALS_SECRET ?? '',
-    sql: `SELECT ${geoid}
-          FROM ${table}
-          WHERE geom && ST_SetSRID(ST_Point(:long, :lat), 4326)
-          LIMIT 1`,
+    sql: `
+      WITH target_row AS (
```
RP-3, line 32: Made this a subquery and explicitly sorted so that at whatever we return from the API is stable.

api/src/geolocate/get.handler.ts
```diff
@@ -64,7 +87,7 @@ export const handler = async (event: {
   let body: GeolocateApiResponse = {};
 
   try {
-    await Promise.allSettled([
+    await Promise.all([
```
RP-3, line 90: Promise.allSettled is [never rejected](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled), so was swallowing errors.

client/src/components/NationwideMap.vue
```diff
@@ -78,7 +78,14 @@ export default defineComponent({
...
     zoomToLongLat() {
-      if (this.geoState?.geoids?.lat != null && this.geoState?.geoids.long != null) {
+      if(this.geoState?.geoids?.pwsId?.bounding_box){
+        const {minLat, minLon, maxLat, maxLon} = this.geoState?.geoids?.pwsId?.bounding_box;
+        this.map?.fitBounds([
+          [minLat, minLon],
+          [maxLat, maxLon]
+        ]);
+      }
+      else if (this.geoState?.geoids?.lat != null && this.geoState?.geoids.long != null) {
```
RP-3, line 88: I assumed it was a good idea to leave the previous `flyTo` behavior around as a fallback.

client/src/components/PredictionPanel.vue
```diff
@@ -86,7 +86,7 @@ export default defineComponent({
       }
       return null;
     },
-    pwsId(): string | null {
+    pwsId(): BoundedGeoDatum | null {
```
RP-3, line 89: I'm unsure if I've changed this in all the right places. Apologies in advance if this is sheer butchery.

api/src/geolocate/get.handler.ts
```diff
@@ -21,24 +20,48 @@ async function getGeoDataForLatLong(
...
+    sql: `
+      WITH target_row AS (
+        SELECT ${geoid} AS id, geom AS geom
+        FROM ${table}
+        WHERE geom && ST_SetSRID(ST_Point(:long, :lat), 4326)
+          ORDER BY id ASC
+          LIMIT 1
+      )
+      SELECT
+        STRING_AGG(id, '|') AS id,
```
kailamjeter, line 40: what is the call to STRING_AGG(id, '|') for? can you leave a comment explaining?

api/src/geolocate/get.handler.ts
```diff
@@ -21,24 +20,48 @@ async function getGeoDataForLatLong(
...
+  const geoID = results.records ? (results.records[0] || undefined) : undefined;
+  if(!geoID) return undefined;
 
-  // TODO(breuch): Consider updating this to all geoids when we support
-  // county, states, etc.
-  return geoids[0];
+  const [rawId, rawBoundingBox] = geoID;
+
+  const boundingBoxVals = rawBoundingBox.stringValue
+  ?.replace('BOX(', '').replace(')', '') // strip out the BOX() wrapper
```
kailamjeter, line 53: these comments are great, but can we also just leave one above this block like 'Query returns bounding box in the format '...'.' And link the postgis documentation?

api/src/geolocate/get.handler.ts
kailamjeter: Great catch, thank you!

client/src/components/NationwideMap.vue
kailamjeter: Agree, thank you!

kailamjeter (approved):

api/src/geolocate/get.handler.ts
RP-3: Done :)

api/src/geolocate/get.handler.ts
RP-3: Done. I made it a bit of an essay but I figure clarity is better.

api/src/geolocate/get.handler.ts
RP-3: :bowtie:

client/src/components/PredictionPanel.vue
beekley: The client should fail to compile if you had missed something, IIUC.

beekley (approved):
Thanks for doing this. Great comments and descriptive PR.

client/src/components/PredictionPanel.vue
RP-3: Perfect. Makes me feel better about it 😅

RP-3: Gotcha. Makes me feel a little better 😅

PR #104, opened 2022-08-17
Landing page styling

kailamjeter:
## Description

Addresses: Styling of landing page

Add sections / new content + styling to landing page. Did not include p1 sections for which implementation details are not finalized (video tutorial + email subscription).

### New

- Moved sections to components
- Added styling prop to landing page section for dynamic styling of sections
- Disabled search button if there is no result
- Moved landing page specific code to landing_page directory

## Testing and Reviewing

ran locally + searched for an address

## Screenshot

https://screenshot.googleplex.com/ZyQjmsx3V6uXhsu


client/src/assets/messages/landing.ts
```diff
@@ -1,34 +1,75 @@
...
-
 /**
  * Strings related to the scorecard-related part of the landing page.
  */
 export class ScorecardSearch {
-  static HEADER = 'Get your drinking water score';
-  static BODY = `Understand the likelihood of your home or community's exposure to lead service lines.`;
-  static CTA_PLACEHOLDER = 'Enter your address';
-  static CTA_BUTTON = 'Get Score';
+  static HEADER = 'How Safe is My Water?';
```
justinebreuch, line 7: Can you make these variables more specific to what they are? We want the content writers to edit this in github directly and I think it'll be easier for them to find things

justinebreuch: Throughout the file

client/src/assets/styles/style_props.ts
```diff
@@ -0,0 +1,65 @@
+// Styles for resources page.
+
+interface Style {
```
justinebreuch, line 3: I don't want to block the CL on this bc I know the pattern was introduced before but I'm worried about defining style props. What I have *started* doing is a map of styling a parent can pass in. The style requirements for each component is going to vary wildly so I don't think defining an interface makes sense here 

https://github.com/BlueConduit/open-data-platform/blob/1307a59ec063fb0559896d59290e8779e6d6ebda/client/src/components/MapLegend.vue#L80

client/src/components/landing_page/ExploreMapSection.vue
```diff
@@ -0,0 +1,68 @@
...
+    LandingPageSection,
+  },
+  data() {
+    return {
+      messages,
+      styles,
+    };
+  },
+  methods: {
+    goToMap(): void {
```
justinebreuch, line 36: I can edit my PR as well but just so we are consistent throughout. Do you want to call routing changes "goTo..." or "navigateTo"

client/src/components/landing_page/ExploreMapSection.vue
```diff
@@ -0,0 +1,68 @@
...
+  width: 920.5px;
+  height: 467.5px;
+  border-radius: 15px;
+
+  background: #FBFBFB;
+  box-shadow: 0 4px 6px 2px rgba(70, 70, 70, 0.25);
+}
+
+.section {
+  padding: 71px 25px;
```
justinebreuch, line 66: This and other pixel values are really specific. The only reason it jumps out at me is generally when this kind of thing happens something has gone awry because there's too many css rules. Generally we want consistent padding on each page

client/src/components/landing_page/LandingPageSection.vue
```diff
@@ -0,0 +1,100 @@
...
+      type: Object as PropType<Messages>,
+      required: true,
+    },
+    styles: {
+      type: Object as PropType<Style>,
+      required: true,
+    },
+  },
+  computed: {
+    cssProps(): Object {
```
justinebreuch, line 43: nit: cssVars for consistency

client/src/components/landing_page/LandingPageSection.vue
```diff
@@ -0,0 +1,100 @@
...
+        '--background-color': this.styles.backgroundColor ?? LandingPageStyles.DEFAULT_STYLE.backgroundColor,
+        '--height': this.styles.height ?? LandingPageStyles.DEFAULT_STYLE.height,
+      };
+    },
+  },
+});
+</script>
+
+<style scoped>
+h1, h2 {
```
justinebreuch, line 58: Cautioning against using h1, h2 headers since you have to overwrite so much css. in global styles theres predefined classes for this so that if we decide to change the h1 for mobile or something it will be applied throughout + less overwrites have to happen with padding and such

client/src/assets/messages/landing.ts
kailamjeter: This isn't possible with the current setup which requires an interface to be passed into the general section component. As discussed offline it may be better to remove the shared component since styling + format of messages differs across sections.

client/src/assets/styles/style_props.ts
kailamjeter: I can change to this but I feel like the complexity of this surpasses the value we get from a shared section component, which really only saves a few lines in html from each component. Let's discuss in standup.

client/src/components/landing_page/ExploreMapSection.vue
kailamjeter: This is what mocks specified, which I agree doesn't make sense written out but was trying to stick as close as possible to that. I'll change it.

client/src/components/landing_page/LandingPageSection.vue
kailamjeter: Didn't realize we had classes for this and Brett left a comment on last PR to the opposite effect, so have been using h1/h2 consistently. Agree it makes sense to use one or the other. I'll remove h1/h2.

client/src/components/landing_page/ExploreMapSection.vue
kailamjeter: I'll switch!

client/src/components/landing_page/LandingPageSection.vue
kailamjeter: Removed component.

client/src/components/landing_page/LandingPageSection.vue
kailamjeter: Switched to shared header classes (additionally added large + xl sizes where needed).

client/src/assets/messages/landing.ts
kailamjeter: Removed use of interface and added more specific names.

client/src/components/landing_page/ExploreMapSection.vue
kailamjeter: Changed to multiples of 4 (copper has a var for 4px). Let me know if these are still too disparate and I'll try to unify.

client/src/assets/styles/style_props.ts
kailamjeter: Removed use of the component so this no longer applies.

justinebreuch:
Discussed offline: Refactor CSS into mixins in a follow-up PR

justinebreuch (approved):

PR #107, opened 2022-08-17
Use env variable for API path.

beekley:
## Description

https://github.com/BlueConduit/open-data-platform/pull/96 removed a hardcoded path to the API, which local development depended on. This moves it to an env variable.

Everyone should add a new line to their `.env` file that points to the API they want to use. You can check the console to see if it's using the right one. E.g.:

```
VUE_APP_API_ENDPOINT=https://beekley.leadout-sandbox.blueconduit.com
```

### New

- `VUE_APP_API_ENDPOINT` variable for local dev. 

### Changed

- API path.

## Testing and Reviewing

My locally-built client works and makes requests to my sandbox.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "serve": "vue-cli-service serve",
+    "serve": "vue-cli-service serve --mode=development",
     "build": "vue-cli-service build",
     "lint": "vue-cli-service lint"
   },
--- a/client/README.md
+++ b/client/README.md
@@ -1,32 +1,37 @@
 # Vue.js app
 
 ## Project setup
+
 ```
 npm install
 ```
 
 ### Compiles and hot-reloads for development
+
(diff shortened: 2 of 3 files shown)


client/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "serve": "vue-cli-service serve",
+    "serve": "vue-cli-service serve --mode=development",
```
kailamjeter, line 6: what does this do?

kailamjeter (approved):

client/package.json
beekley: [Despite the docs](https://cli.vuejs.org/guide/mode-and-env.html#modes), `serve` here was using `.env.production`. I had to add this to force it to use my `.env` file. I'd describe that as a comment but json doesn't support comments.

kailamjeter: Interesting, thanks!

PR #108, opened 2022-08-17
Add new logos + colors

beekley:
Slight style changes to try to match [the mocks](https://www.figma.com/file/0LGxjZnPD3N0dCqHv8LRr6/NW-Map-Design-System-WIP?node-id=452%3A5994) more (though the goal in this PR isn't to match the CSS exactly).

### New

- LeadOut logo and favicon.

### Changed

- Points colors to the [Copper design tokens](https://github.com/BlueConduit/copper/blob/main/packages/copper/scss/01_settings/_design-tokens.scss). These don't have the new colors yet, so we'll switch over when they're added. 

## Testing and Reviewing

Screen Shot 2022-08-17 at 3 06 12 PM


client/public/index.html
```diff
@@ -1,21 +1,21 @@
...
+    <script src="https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-geocoder/v5.0.0/mapbox-gl-geocoder.min.js"></script>
+    <link href="https://api.mapbox.com/mapbox-gl-js/v2.8.2/mapbox-gl.css" rel="stylesheet" />
+    <script src="https://api.mapbox.com/mapbox-gl-js/v2.8.2/mapbox-gl.js"></script>
+    <link rel="icon" href="<%= BASE_URL %>favicon.ico" />
     <title>LeadOut</title>
   </head>
   <body>
     <noscript>
-      <strong>We're sorry but LeadOut doesn't work properly without JavaScript enabled. Please enable it to continue.</strong>
+      <strong
```
kailamjeter, line 15: can you change the formatting to 

<strong>
    ...
</strong>

this default line wrapping style for html is strange

client/src/components/NavigationBar.vue
```diff
@@ -49,23 +34,19 @@ export default defineComponent({
...
-        '--height': `${this.height + 'px'}`,
-      };
-    },
-  },
 });
 </script>
 
-<style scoped>
+<style scoped lang="scss">
+@use '@blueconduit/copper/scss/01_settings/design-tokens' as dt;
```
kailamjeter, line 41: nit: just added these also and found the 'dt' prefix to be annoying / repetitive - can you import instead of use and remove it? 

if you disagree and think we should keep prefix I can add it, but we should be consistent

kailamjeter (approved):

justinebreuch (approved):
Nice!

client/src/components/NavigationBar.vue
beekley: SGTM and done.

client/public/index.html
beekley: Haha yes this line wrapping was bizarre.

PR #109, opened 2022-08-18
Add Google Analytics to non-local envs.

beekley:
## Description

We decided to use GA because it's free. This adds the snippet so we can start gathering data. The snippet is only active in `production`, which to Vue means any time we run `npm run build`. At least, this is used in dev and prod.

### New

- GA snippet.

The [dashboard](https://analytics.google.com/analytics/web/?authuser=3#/p327821062/reports/intelligenthome) shows data from my sandbox:

image


jbhammon (approved):
The templating pieces look correct to avoid collecting data during local development. 

The GA code looks familiar from past projects, but I'm just assuming that part is correct.

PR #111, opened 2022-08-18
About page

kailamjeter:
## Description

Addresses: Adds content for about us page

Adds 'About us' page view. Note that values for css in mocks seemed a bit off when applied so went on what looked best to me -- open to comments on how things look!

## Testing and Reviewing

ran locally + navigated to the page using footer

## screenshot
Screen Shot 2022-08-19 at 10 14 26 AM


client/src/views/AboutUsView.vue
```diff
@@ -0,0 +1,60 @@
...
+  height: 100%;
+  background-color: $light-blue-800;
+  background-image: url('~@/assets/media/about-us-image.png');
+  background-size: cover;
+  background-repeat: no-repeat;
+  color: $white;
+}
+
+.h1-header {
+  width: 169 * $spacing-xs;
```
beekley, line 45: nit: multiplying by a factor here seems to defeat the purpose of using a common value. Would [`max-width`](https://github.com/BlueConduit/copper/blob/main/packages/copper/scss/01_settings/_design-tokens.scss#L55) work instead?

client/src/views/AboutUsView.vue
```diff
@@ -0,0 +1,60 @@
...
+}
+
+.h1-header-xxl {
+  font-weight: 600;
+  line-height: 15 * $spacing-xs;
```
beekley, line 50: nit: [`spacing-xl`](https://github.com/BlueConduit/copper/blob/main/packages/copper/scss/01_settings/_design-tokens.scss#L52) is about the same as this value.

beekley (approved):
LGTM. If Copper doesn't have good values to use, I'd rather just use a literal value (e.g. `3.75rem` than coerce something from Copper (e.g. `15 * spacing-xs`) so we don't have to look up a file and do math to figure out what it means.

We can always copy it over to Copper in the future for code reuse.

client/src/assets/messages/about_us.ts
```diff
@@ -0,0 +1,13 @@
+export class AboutUs {
+  static HEADER = 'About us';
+  static BODY_1 =
+    'Leadout is a tool created by BlueConduit. BlueConduit is a water analytics company that has ' +
+    'developed cutting-edge, predictive machine learning software to locate lead service lines, ' +
+    'empowering local officials and their engineering partners with the information to ' +
+    'efficiently remove those pipes.';
+  static BODY_2 =
```
justinebreuch, line 8: nit: make these more descriptive var names

client/src/assets/styles/global.scss
```diff
@@ -1,9 +1,19 @@
...
+$light-blue-800: #0576BD;
+
 /** Typography */
 
 .h1-header {
   font-size: 24px;
 }
 
+.h1-header-xxl {
+  font-size: 80px;
```
justinebreuch, line 14: This is reall big - even bigger than the logo I think. I think it's fine to keep for now I'm just adding a task to lead out to standardize these fonts. Can you do the same thing you did elsewhere 4px * xxx?

client/src/views/AboutUsView.vue
justinebreuch: Is this the same xxl as is in the global styles?

client/src/views/AboutUsView.vue
```diff
@@ -0,0 +1,60 @@
...
+  text-transform: capitalize;
+}
+
+.text {
+  display: flex;
```
justinebreuch, line 55: nit: container-column class or mixin

justinebreuch (approved):

client/src/views/AboutUsView.vue
kailamjeter: Removed this but changed elsewhere

client/src/assets/styles/global.scss
kailamjeter: Made 56px instead (what I previously added as h1-header-xl). Will do a followup PR to remove from landing page styles I submitted but yes will fix here. Thanks!

client/src/views/AboutUsView.vue
kailamjeter: Done

client/src/views/AboutUsView.vue
kailamjeter: Changed to max-width and just a px value. Left a TODO to add a var for these values. Thanks!

client/src/assets/messages/about_us.ts
kailamjeter: Done

PR #112, opened 2022-08-19
CSS clean up + show error when no prediction came through

justinebreuch:
## Description

Addresses: Add error handling on any API failures

This is just covering all failures right now in the prediction panel. See screenshot. 

Also a bunch of css cleanup to conform to copper.

## Testing and Reviewing
Right now my API is always failing so this was a good time to do the PR:

Screen Shot 2022-08-19 at 11 22 59 AM
 checklists for how someone can review this PR to verify it's working as intended_


client/src/views/ScorecardView.vue
```diff
@@ -72,9 +69,12 @@ export default defineComponent({
...
 </script>
 
-<style scoped>
+<style scoped lang='scss'>
+@import '../assets/styles/global.scss';
+@import '@blueconduit/copper/scss/01_settings/design-tokens';
+
 .actions-to-take {
-  background-color: #E1F5FE;
+  box-shadow: 0px 0px 8px #757575;
```
beekley, line 77: Would one of [the shadows from Copper](https://github.com/BlueConduit/copper/blob/main/packages/copper/scss/01_settings/_design-tokens.scss#L66) work here?

beekley (approved):

client/src/assets/styles/global.scss
```diff
@@ -1,3 +1,8 @@
+@use '@blueconduit/copper/scss/01_settings/design-tokens' as dt;
```
kailamjeter, line 1: nit can you do as * instead so we don't have to prefix with dt.

client/src/assets/styles/global.scss
```diff
@@ -71,6 +72,21 @@ $navy-blue: #2553A0;
...
+  display: flex;
+}
+
+@mixin container-column {
+  display: flex;
+  flex-direction: column;
+}
+
+@mixin centered-container {
+  @include container-column;
```
kailamjeter, line 86: can we keep this agnostic of direction?

kailamjeter (approved):

client/src/assets/styles/global.scss
justinebreuch: Done!

client/src/assets/styles/global.scss
justinebreuch: That's a good idea

client/src/views/ScorecardView.vue
justinebreuch: Obsolete!

PR #113, opened 2022-08-19
Get full URL from window instead of env var.

beekley:
## Description

Just so we don't need to use an env var (which was broken in dev).

Also this does some linting changes. I don't know what causes those and I've spent too much time trying to get it consistent across devs' environments.


client/src/views/ScorecardView.vue
```diff
@@ -74,7 +79,7 @@ export default defineComponent({
 
 <style scoped>
 .actions-to-take {
-  background-color: #E1F5FE;
+  background-color: #e1f5fe;
```
kailamjeter, line 82: nit: know you didn't add this but can you change this (and above if possible, not sure about the scss in js thing) to use the color constants defined in global.scss?

kailamjeter (approved):

justinebreuch (approved):

client/src/views/ScorecardView.vue
beekley: done.

PR #115, opened 2022-08-19
Improve scorecard map load time

beekley:
## Description

Addresses: [Restrict pan and zoom in scorecard](https://tables.area120.google.com/table/bX_kMNBW8LA35n6l5Hz_co/row/bjMXlJUp0Ze5_iQ-ZA2kwb)

Problem statements:
* Loading the map on a scorecard takes ~5 seconds to render, even when all of the tiles are cached. 
* Panning in the scorecard is slow and stuttery (it uses 100% of my Macbook Pro CPU).

The scorecard view is meant to just show data about the given entity (e.g. parcel, zip), while the map puts that data in context. Users aren't expected to use this view as an entry point for browsing the whole set of nationwide map data.

This PR disables animations and user zoom/pan interaction with the scorecard map. 

TODO: look into simplifying mapbox JS altogether ([context](https://lslr.slack.com/archives/C03D5199TLG/p1660952368390959))

### New

- _Any new things_

### Changed

- _Any changed things_

### Removed

- _Any removed things_

## Testing and Reviewing

TODO: gather data.

See it live in my sandbox: https://beekley.leadout-sandbox.blueconduit.com/scorecard/39.965903,-75.1811125

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -25,7 +25,7 @@
         "@babel/core": "^7.12.16",
         "@babel/eslint-parser": "^7.12.16",
         "@types/mapbox__mapbox-gl-geocoder": "^4.7.2",
-        "@types/mapbox-gl": "^2.7.4",
+        "@types/mapbox-gl": "^2.7.5",
         "@types/vue-select": "^3.16.0",
         "@typescript-eslint/eslint-plugin": "^5.4.0",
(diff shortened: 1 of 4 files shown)


beekley:
I might come back to this, but not right now and likely not with this exact approach. So I'm closing this PR.

PR #119, opened 2022-08-22
Limit geocoding to postcode + address types in search lead status context

kailamjeter:
## Description

Limits geocode component to certain address and zip code in landing page and lead status page. This still accepts all types in the map page and fixes map search which was not hooked up w/ v-model before.

## Testing and Reviewing

ran locally and tried inputting different geo types in the different contexts.

### DIFF
--- a/client/src/components/GeocoderInput.vue
+++ b/client/src/components/GeocoderInput.vue
@@ -46,7 +46,7 @@ export default defineComponent({
     this.geocoder.setPlaceholder(this.placeholder);
     
     if (this.acceptedTypes != null) {
-      this.geocoder.setTypes(this.acceptedTypes.join());
+      this.geocoder.setTypes(this.acceptedTypes.join(','));
     }
 
     // TODO: replace 'any' here with a meaningful type.
--- a/client/src/components/landing_page/ScorecardSearch.vue
+++ b/client/src/components/landing_page/ScorecardSearch.vue
@@ -2,7 +2,7 @@
   <div class='scorecard-search'>
     <GeocoderInput class='geocoder'
                    :placeholder='this.placeholder'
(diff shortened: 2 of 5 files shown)


client/src/components/GeocoderInput.vue
```diff
@@ -40,6 +44,10 @@ export default defineComponent({
   mounted() {
     this.geocoder.addTo('.geocoder');
     this.geocoder.setPlaceholder(this.placeholder);
+    
+    if (this.acceptedTypes != null) {
+      this.geocoder.setTypes(this.acceptedTypes.join());
```
beekley, line 49: This looks weird without a comma in the `join` call, since I assume `setTypes` takes a comma-separated list. But if it works then please ignore.

beekley (approved):

client/src/components/landing_page/ScorecardSearch.vue
```diff
@@ -2,6 +2,7 @@
   <div class='scorecard-search'>
     <GeocoderInput class='geocoder'
                    :placeholder='this.placeholder'
+                   :accepted-types='acceptedTypes'
```
justinebreuch, line 5: nit: acceptedTypes

justinebreuch (approved):

client/src/components/GeocoderInput.vue
kailamjeter: It does, and join does by default but this is a good point, I'll add it just for clarity and so if join behavior changes this won't break.

client/src/components/landing_page/ScorecardSearch.vue
kailamjeter: done!

PR #121, opened 2022-08-22
Update raw values for scorecard summary factors

justinebreuch:
## Description

Addresses: Finalize scorecard content + styling

Updates some styling on the scorecard page to closer align with other pages. 

Contextualizes numbers in scorecard factors for now. These numbers will be run through UXR testing to determine whether we should show them at all. 

### New
Formats the demographic numbers to text rather than raw numbers. See screenshot. 

## Testing and Reviewing
Screen Shot 2022-08-22 at 4 04 15 PM

### DIFF
--- a/client/src/components/ScorecardSummaryPanel.vue
+++ b/client/src/components/ScorecardSummaryPanel.vue
@@ -147,7 +147,6 @@ export default defineComponent({
 
 .container-column {
   max-width: 12 * $spacing-xl;
-  gap: $spacing-lg;
   padding: $spacing-lg;
 }
 </style>
--- a/client/src/components/ScorecardSummaryPanel.vue
+++ b/client/src/components/ScorecardSummaryPanel.vue
@@ -6,7 +6,8 @@
           zipCode) }}
         </div>
       </div>
-      <div> {{ ScorecardSummaryMessages.SCORECARD_SUMMARY_PANEL_SUBHEADER }}
+      <div class='subheader'>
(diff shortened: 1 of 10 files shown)


kailamjeter:
I know content isn't finalized but for now can you change that first sentence to:

'In addition to information from your water system, your score is also based on these statistics for your zip code:"

Just to explicitly mention that these numbers are for zip.

client/src/components/ScorecardSummaryPanel.vue
```diff
@@ -63,13 +73,47 @@ export default defineComponent({
   },
   computed: {
     homeAgeComparison(): string | null {
-      return this.formatNumber(this.demographicDataState?.data?.averageHomeAge);
+      const homeAge = this.roundNumber(this.demographicDataState?.data?.averageHomeAge);
```
kailamjeter, line 76: nit can we check if this is null before rounding and just return?

same below for income

kailamjeter (approved):

client/src/components/ScorecardSummaryPanel.vue
justinebreuch: Round number does that - it checks if it's null and just returns null in that case

justinebreuch:
> I know content isn't finalized but for now can you change that first sentence to:
> 
> 'In addition to information from your water system, your score is also based on these statistics for your zip code:"
> 
> Just to explicitly mention that these numbers are for zip.

Good idea. Updated

PR #122, opened 2022-08-22
tiny styling changes

kailamjeter:
Changes button hover / disabled color and makes change to about us change text alignment.

### DIFF
--- a/client/src/views/AboutUsView.vue
+++ b/client/src/views/AboutUsView.vue
@@ -41,10 +41,6 @@ export default defineComponent({
   color: $white;
 }
 
-.h1-header {
-  text-align: start;
-}
-
 .text {
   gap: $spacing-lg;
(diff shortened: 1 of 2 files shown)


client/src/views/AboutUsView.vue
```diff
@@ -41,9 +41,13 @@ export default defineComponent({
   color: $white;
 }
 
+.h1-header {
+  text-align: start;
+}
+
```
justinebreuch, line 47: If you merge main, do you still need this? I removed centered text out of h1-header into its own mixin

justinebreuch (approved):

client/src/views/AboutUsView.vue
kailamjeter: Nope - thanks!

PR #123, opened 2022-08-23
Add slack chatbot to notify on alerts

beekley:
## Description

Addresses: [Monitoring](https://tables.area120.google.com/table/bX_kMNBW8LA35n6l5Hz_co/row/bYcpg5d5BPEdlvjUIGh_gZ)

We currently have no monitoring. After each deployment (including to their own sandboxes), devs have to look in the logs in the AWS console to see if it succeeded.

This PR adds a chatbot and one integration to demonstrate how monitoring can work:

1. A stack defines an alert for some metric (e.g. lambda error rate).
2. The monitoring stack subscribes to an SNS topic associated with that alert.
3. The chatbot posts whenever an alert is written to that topic.

However, I don't love this CDK setup since it makes the monitoring stack dependent on all other stacks. I've had trouble updating other stacks with errors like `... cannot be deleted as it is in use by beekley-OpenDataPlatformMonitoring`. I'll look into making it more independent in a future PR.

### New

- Monitoring stack with a slack chatbot.
- Alert for schema update errors.

### Changed

- Fixed a real schema.sql error.

## Testing and Reviewing

My schema was failing to update for [real reasons](https://us-east-2.console.aws.amazon.com/cloudwatch/home?region=us-east-2#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fbeekley-OpenDataPlatformD-RootSchemahandler909455B-xEMDumOKKiDN/log-events/2022$252F08$252F23$252F$255B$2524LATEST$255D6712eedfeb2e4ac09fe42f67adaf927c), and the chatbot [successfully alerted](https://lslr.slack.com/archives/C03V1FX7KC1/p1661269502351689) in [#leadout-sandbox-notifications](https://app.slack.com/client/TJTFN34NM/C03V1FX7KC1).

PR #125, opened 2022-08-23
Refactor monitoring to its own CDK app

beekley:
## Description

https://github.com/BlueConduit/open-data-platform/pull/123 added a slack chatbot integration. However, there can only be one chatbot with a given config, so our sandboxes failed to deploy when one already existed.

This PR moves the monitoring stack to its own CDK app, which is deployed separately from the rest of the stacks. Then, just one person will deploy it for sandbox (currently me) and the other sandboxes can reference it through its SNS topic.

### New

- Monitoring CDK app, deployed with `cdk run deploy-monitoring`.

### Changed

- Moved monitoring stack to the new app.

## Testing and Reviewing

My sandbox [deploys](https://beekley.leadout-sandbox.blueconduit.com/), even when there's an existing [chatbot](https://us-east-2.console.aws.amazon.com/chatbot/home?region=us-east-2#/chat-clients/slack/workspaces/TJTFN34NM/configurations/SANDBOX-OpenDataPlatform-channel-config).


beekley:
@arockett This PR implements what you described in the last one about exposing an SNS Topic instead of the chatbot itself. Seems like a good approach!

kailamjeter (approved):

PR #131, opened 2022-08-24
Show build info in the client.

beekley:
## Description

This PR adds a log statement in the client that displays information about the current build. This allows anyone to see how recent the code is, including a link to the git commit it was built from, if present.

I considered displaying the ID somewhere in the client GUI itself, but this is faster and only visible to those who look for it. I'm open to dropping it in the corner somewhere too.

### New

- Console log statement in the client.

## Testing and Reviewing

I added `CODEBUILD_SOURCE_REPO_URL=testURL` to my `.env` file and see the following in my console:

image

We'll know for sure if the git URL works once this is deployed to dev.

### DIFF
--- a/client/public/index.html
+++ b/client/public/index.html
@@ -1,6 +1,5 @@
 <!DOCTYPE html>
-<!-- Create timestamp at build-time. -->
-<html lang="" data-build-timestamp-utc="<%= new Date().toISOString() %>">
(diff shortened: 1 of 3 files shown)


beekley:
Not ready yet.

kailamjeter (approved):

PR #133, opened 2022-08-25
Lslr section

kailamjeter:
## Description

Adds LSLR section to scorecard page.

This requires adding city to the  API response and populating with the correct value on the insert lambda.

Additionally adds geoType to URL when searching from landing page. This triggers a parcel lookup when the user searches by address.

### New

- LSLR section component

## Testing and Reviewing

Deployed API changes and ran locally to test full stack. Tested searching both an address with a LSLR link and without to test the toggling.

### DIFF
--- a/cdk/src/open-data-platform/data-plane/data-import/data-import-stack.ts
+++ b/cdk/src/open-data-platform/data-plane/data-import/data-import-stack.ts
@@ -46,7 +46,7 @@ export class DataImportStack extends Construct {
 
     const { cluster, credentialsSecret } = props;
 
-    const s3BucketWithDataFiles = new s3.Bucket(this, 'open_data_platform_static_files');
+    const s3BucketWithDataFiles = new s3.Bucket(this, `${id}-open_data_platform_static_files`);
 
     // Allow reads to all S3 buckets in account.
     const s3GetObjectPolicy = new iam.PolicyStatement({
--- a/cdk/src/open-data-platform/data-plane/data-import/write-parcels-data.handler.ts
+++ b/cdk/src/open-data-platform/data-plane/data-import/write-parcels-data.handler.ts
@@ -64,7 +64,7 @@ function getTableRowFromRow(row: any): SqlParametersList {
   return (
     new ParcelsTableRowBuilder()
       .address(properties.address)
-      .city(properties.city ?? '')
+      .city('toledo') // TODO: change this if we load other city parcel data.
       // TODO: Change to real properties once we have them.
       .publicLeadLowPrediction(getValueOrDefault(properties.y_score_pub))
       .publicLeadHighPrediction(getValueOrDefault(properties.y_score_pub))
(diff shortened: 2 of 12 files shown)


cdk/src/open-data-platform/data-plane/data-import/data-import-stack.ts
```diff
@@ -46,7 +46,7 @@ export class DataImportStack extends Construct {
 
     const { cluster, credentialsSecret } = props;
 
-    const s3BucketWithDataFiles = new s3.Bucket(this, 'open_data_platform_static_files');
+    const s3BucketWithDataFiles = new s3.Bucket(this, `${id}-open_data_platform_static_files`);
```
kailamjeter, line 49: @justinebreuch not sure if we want to add this permanently but left for now -- I'll remove if you prefer

client/src/components/LslrSection.vue
```diff
@@ -0,0 +1,61 @@
...
+import { defineComponent, PropType } from 'vue';
+import { ScorecardMessages as messages } from '@/assets/messages/scorecard_messages';
+import { City } from '@/model/states/model/geo_data';
+
+export const LSLR_CITY_LINKS: Map<City, string> = new Map<City, string>([
+  [City.toledo, 'https://toledo.oh.gov/residents/water/lead-service-lines/customer-side'],
+]);
+
+/**
+ * Lslr section component.
```
beekley, line 25: nit: could you define what LSLR means here or somewhere?

cdk/src/open-data-platform/data-plane/data-import/write-parcels-data.handler.ts
```diff
@@ -4,14 +4,15 @@ import { ParcelsTableRowBuilder } from '../model/parcels-table';
 import { geoJsonHandlerFactory } from './handler-factory';
 
 const SCHEMA = 'schema';
+const CITY = 'toledo';
```
beekley, line 7: A suggestion for a future PR: we could create a handler for each city by iterating through the city enum in `/client/src/model/states/model/geo_data` and passing its value to the hander as an argument.

beekley (approved):

cdk/src/open-data-platform/data-plane/data-import/data-import-stack.ts
justinebreuch: No this is great thanks!

cdk/src/open-data-platform/data-plane/data-import/write-parcels-data.handler.ts
kailamjeter: Good idea! I'm not sure you can use client variables here but we could create an identical one in data import stack to use.

client/src/components/LslrSection.vue
kailamjeter: Done!

PR #135, opened 2022-08-25
Scorecard search

kailamjeter:
## Description

Adds lat/long to URL from search on scorecard page and clears state between pages. Also removes duplicate marker.

### New

- A few reducers to clear the state when a user switches pages.

## Testing and Reviewing

ran locally, searched different results from the scorecard, and toggled between pages.

### DIFF
--- a/client/src/model/slices/demographic_data_slice.ts
+++ b/client/src/model/slices/demographic_data_slice.ts
@@ -1,15 +1,9 @@
 import { createSlice, PayloadAction } from '@reduxjs/toolkit';
 import { ApiClient } from '@/api/api_client';
 import { AppDispatch } from '@/model/store';
-import { LeadData } from '@/model/states/model/lead_data';
-import { LeadDataState } from '@/model/states/lead_data_state';
 import { DemographicDataState } from '@/model/states/demographic_data_state';
 import { DemographicData } from '@/model/states/model/demographic_data';
 import { GeographicLevel } from '@/model/data_layer';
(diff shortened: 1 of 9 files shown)


client/src/model/slices/demographic_data_slice.ts
```diff
@@ -25,6 +29,11 @@ const demographicDataSlice = createSlice({
     getDemographicsError(state: DemographicDataState, action) {
       console.log(`Error fetching demographic data: ${state} ${action}`);
     },
+    demographicDataCleared(state: GeoDataState, action) {
```
justinebreuch, line 32: Could you try one thing both here and in other slicers. 
```
    getDemographicsSuccess(state: DemographicDataState, action: PayloadAction<DemographicData>) {
      return {
        ...state,
        data: { ...action.payload },
        status: { status: Status.pending },
      };
    },

...

export const clearDemographicData = () => {
  return async (dispatch: AppDispatch) => {
    dispatch(getDemographicsSuccess({}));
  };
};
```

justinebreuch (approved):

client/src/model/slices/demographic_data_slice.ts
kailamjeter: Oh that's a good point, will give that a try!

kailamjeter: As discussed offline will keep separate since we don't want to retain existing state info on clear.

PR #136, opened 2022-08-26
Refactor app state to map state 

justinebreuch:
## Description

Addresses: Remove app state in favor of map state 

Refactor out app state. Create a new redux state for map state. Anytime a layer changes, an event gets emitted. Nationwide map listens to that event. 

### New

The map state, map data state, reducers to change both zoom and current data layer. 

### Changed

Anywhere that used app state now has map state

### Removed

App state, injection keys.

## Testing and Reviewing

in UI - I tried testing all known functionality


client/src/components/NationwideMap.vue
```diff
@@ -67,13 +66,30 @@ export default defineComponent({
         '--height': this.height,
       };
     },
+    currentDataLayerId() {
+      return this.mapState?.mapData?.currentDataLayerId;
+    },
+    visibleLayer() {
+      return this.possibleLayers?.find(l =>
+        this.map?.getLayer(l.styleLayer.id) != null &&
```
kailamjeter, line 74: why do we need this check?

kailamjeter: Looking below, is this because we can have layers in ALL_DATA_LAYERS which aren't in the map?

client/src/components/NationwideMap.vue
```diff
@@ -112,21 +128,36 @@ export default defineComponent({
     /**
      * Sets the visibility of the layer with the given styleLayerId.
      */
-    setDataLayerVisibility(styleLayerId: string, visible: boolean): void {
+    setDataLayerVisibility(layerId: string): void {
```
kailamjeter, line 131: lets change this to setDataLayerVisible if it just sets layerId visible now

client/src/components/NationwideMap.vue
```diff
@@ -194,14 +220,14 @@ export default defineComponent({
      */
     setUpInteractionHandlers(): void {
       if (this.map == null) return;
-      for (const layer of this.state.dataLayers) {
+      for (const layer of this.possibleLayers) {
         // Use MapBox's custom click handler, which takes the style layer that we
         // want to set up a handler for as a parameter.
         this.map?.on('click', layer.styleLayer.id, async (e: MapLayerMouseEvent): Promise<void> => {
-          if (e.features != undefined) {
+          if (e.features != null) {
```
kailamjeter, line 227: should we check for null and undefined?

client/src/components/NationwideMap.vue
```diff
@@ -281,63 +307,68 @@ export default defineComponent({
...
     // Listen for changes to lat/long to update map location.
     'geoState.geoids': function() {
       const lat = this.geoState?.geoids?.lat;
       const long = this.geoState?.geoids?.long;
 
       if (lat != null && long != null) {
-        const marker = new mapboxgl.Marker({
+        // Remove the old marker.
+        if (this.marker != null) {
+          this.marker.remove();
```
kailamjeter, line 363: the only change i made was moving this outside of the if

client/src/router.ts
```diff
@@ -80,6 +81,13 @@ const router = createRouter({
   },
 });
 
+/// Remove layer query from scorecard page.
```
kailamjeter, line 84: is it difficult to prevent it from being added at all?

kailamjeter (approved):

client/src/components/NationwideMap.vue
justinebreuch: If there's ever a state where the map does not have anything visible, then it should check the currentDataLayer

client/src/components/NationwideMap.vue
justinebreuch: Done and combined with toggle data layer since it really only had one extra line

client/src/router.ts
justinebreuch: It happens through the nav bar through the native router-links. From the searches I did, I couldn't find much about it. I also removed the ? From the scorecord path but that didn't do anything. If I don't do this, the about page is also http://localhost:8080/about?layer=lead-service-lines-by-water-system

client/src/components/NationwideMap.vue
justinebreuch: the null check actually does both!

client/src/components/NationwideMap.vue
justinebreuch: oo good point - swapped it

client/src/App.vue
```diff
@@ -63,11 +50,16 @@ export default defineComponent({
...
-        // TODO: Pass real geo type.
         dispatch(queryLatLong(lat, long, geoType));
       }
 
+      // Check whether router has a param with the layer to show on the map.
+      // Otherwise, default to water systems.
+      const currentDataLayerId = this.mapState?.mapData?.dataLayers?.find(
+        (l) => layerId == l,
+      );
+      dispatch(setCurrentDataLayer(currentDataLayerId ?? MapLayer.LeadServiceLineByWaterSystem));
```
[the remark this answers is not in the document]
justinebreuch: Setting the url for the scorecard won’t do anything since we have to remove it so we need to update the state not through the URL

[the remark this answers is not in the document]
justinebreuch: Oh and map auto-updates its URL

PR #137, opened 2022-08-26
Fix zoom out

kailamjeter:
## Description

Fixes clear search bug (would zoom to 0,0) and adds call to fetch water system data for addresses (previously only fetched parcel data)

## Testing and Reviewing

ran locally, searched for parcel without data, cleared search

### DIFF
--- a/client/src/components/NationwideMap.vue
+++ b/client/src/components/NationwideMap.vue
@@ -362,6 +362,14 @@ export default defineComponent({
           this.marker.setLngLat([parseFloat(long), parseFloat(lat)]).addTo(this.map);
         }
         this.zoomToLongLat();
+      } else {
+        // Reset map to full view of U.S. when geo IDs are cleared.
+        if (this.map?.isStyleLoaded()) {
+          this.map?.jumpTo({
+            center: this.center,
+            zoom: 4,
+          });
+        }
       }
     },
   },
--- a/client/src/components/NationwideMap.vue
+++ b/client/src/components/NationwideMap.vue
@@ -26,6 +26,7 @@ const VISIBILITY = 'visibility';
 const VISIBLE = 'visible';
 
 const PARCEL_ZOOM_LEVEL = 12;
+const DEFAULT_ZOOM_LEVEL = 4;
 
 /**
  * A browsable map of nationwide lead data.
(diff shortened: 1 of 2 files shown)


client/src/components/NationwideMap.vue
```diff
@@ -345,6 +345,14 @@ export default defineComponent({
           this.marker.setLngLat([parseFloat(long), parseFloat(lat)]).addTo(this.map);
         }
         this.zoomToLongLat();
+      } else {
+        // Reset map to full view of U.S. when geo IDs are cleared.
+        if (this.map?.isStyleLoaded()) {
+          this.map?.jumpTo({
+            center: this.center,
+            zoom: 4,
```
justinebreuch, line 353: nit: Make zoom constant it's used twice in this file now

client/src/components/NationwideMap.vue
```diff
@@ -323,20 +323,20 @@ export default defineComponent({
     },
     // Listens to query param to toggle layers.
     routerLayer: function(newLayer: string) {
-      if (newLayer != null) {
+      if (newLayer != null && this.map?.isStyleLoaded()) {
```
justinebreuch, line 326: Don't even need this watcher since app.vue dispatches an event to change the map's current layer

justinebreuch (approved):

client/src/components/NationwideMap.vue
```diff
@@ -323,20 +323,20 @@ export default defineComponent({
...
         this.setDataLayerVisibility(newLayer, true);
       }
     },
     // Listen for changes to lat/long to update map location.
     'geoState.geoids': function() {
+      if (this.marker) {
+        this.marker.remove();
```
justinebreuch, line 333: You don't know that this marker exists yet do you?

justinebreuch: oh this is checking if it exists I see!

client/src/components/NationwideMap.vue
kailamjeter: Done

client/src/components/NationwideMap.vue
kailamjeter: Removed!

client/src/components/NationwideMap.vue
kailamjeter: Changed to what you had anyway - it's more clear!

PR #138, opened 2022-08-27
Cluster water_systems data by its spatial index

RP-3:
## Description

Addresses: [Complete low-hanging-fruit DB optimizations for tile serving](https://tables.area120.google.com/table/bX_kMNBW8LA35n6l5Hz_co/row/blYIQsdPnww7YU8xsoy_aD)

- '[Clustering](https://www.postgresql.org/docs/current/sql-cluster.html)' physically sorts the bytes on disk in a table
- The effect of this is that reading data from an index sequentially becomes faster because you need fewer disk seeks
- In our case (serving tiles), the polygons for each tile are by-definition close to each other, since they're on the same tile, making this a no-brainer
- The effect of this is that a slow tile fetch (e.g., z, x, y, = 6, 14, 25) gets 10-20% faster.
- Including the bbox index:
  - Exec time before: 1103ms, 1139ms, 1130ms
  - Exec time after: 838ms, 930ms, 831ms
- This PR updates the water_systems import to cluster the table by its spatial index at the end of every import
- Clustering, according to my sandbox test, takes about 2.5 seconds. The table is locked for this period and cannot be accessed. 

### New

- The water-systems import now clusters the table when it has finished writing the rows to disk.

### Changed

- None

### Removed

- None

## Testing and Reviewing

- Run `cdk deploy rsrogers-OpenDataPlatformFrontend --profile=sandbox`, and then run the water-systems import lambda. 
- Observe the effect by just running `CLUSTER water_systems_geom_idx ON water_systems; ` on your db.

### DIFF
--- a/cdk/src/open-data-platform/data-plane/data-import/write-water-systems-data.handler.ts
+++ b/cdk/src/open-data-platform/data-plane/data-import/write-water-systems-data.handler.ts
@@ -1,5 +1,5 @@
 import { RDSDataService } from 'aws-sdk';
-import { BatchExecuteStatementRequest, SqlParametersList } from 'aws-sdk/clients/rdsdataservice';
+import { BatchExecuteStatementRequest, ExecuteStatementRequest, SqlParametersList } from 'aws-sdk/clients/rdsdataservice';
 import { geoJsonHandlerFactory } from './handler-factory';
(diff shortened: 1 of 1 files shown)


justinebreuch (approved):

kailamjeter (approved):
Nice!

PR #139, opened 2022-08-27
Drop smaller polygons at distant zoom levels

RP-3:
## Description

Addresses: [Complete low-hanging-fruit DB optimizations for tile serving](https://tables.area120.google.com/table/bX_kMNBW8LA35n6l5Hz_co/row/blYIQsdPnww7YU8xsoy_aD)

- Some queries fail with the error `Database returned more than the allowed response size limit`
- This PR addresses the error by filtering out smaller polygons at farther zoom levels
- This has a negligible effect on query performance
- Again, since we can't alter tables by re-running schema.sql, you'll have to run the following manually after this is merged:

```sql
ALTER TABLE public.water_systems ADD COLUMN approx_area_sq_km REAL;
UPDATE public.water_systems SET approx_area_sq_km = ST_Area(ST_Transform(geom, 4326)) / 1000000;
CREATE INDEX ON water_systems USING BTREE (approx_area_sq_km);
```

### New

- None

### Changed

- Queries at the following zoom levels only return polygons with square-kilometer sizes greater than specified below
```
zoom | min_area_sq_km
-----+---------------
 <5  | Infinity
 5   | 100
 7   | 10
 8   | 5
 >8  | 0
```

### Removed

- None

## Testing and Reviewing

- Set `VUE_APP_DEFAULT_TILESERVER_HOST=rsrogers.leadout-sandbox.blueconduit.com` in your `.env`
- Scroll around the nationwide map. You shouldn't notice water systems obviously missing, and you also shouldn't see any failed requests in the network tab on the dev console. 
- Check that schema.sql doesn't freak out [here](https://us-east-2.console.aws.amazon.com/cloudwatch/home?region=us-east-2#logsV2:log-groups/log-group/$252Faws$252Flambda$252Frsrogers-OpenDataPlatform-RootSchemahandler909455B-PWnu0M0oqt3N/log-events/2022$252F08$252F27$252F$255B$2524LATEST$255De8d9547083ba464db6fa62c8d0fc8371)
- Delete and reimport water systems and check that it worked.

### DIFF
--- a/cdk/src/open-data-platform/data-plane/schema/schema.sql
+++ b/cdk/src/open-data-platform/data-plane/schema/schema.sql
@@ -428,7 +428,19 @@ BEGIN
                         SUM(w.service_connections_count)       AS service_connections_count,
                         SUM(w.population_served)               AS population_served
                  FROM water_systems w
-                 WHERE ST_Transform(w.bbox, 3857) && ST_TileEnvelope(z, x, y)
+                 WHERE
+                    ST_Transform(w.bbox, 3857) && ST_TileEnvelope(z, x, y) AND
+                    -- Require nearer zoom levels to see smaller polygons. Cutoffs are arbitrary.
+                    -- Necessary because some tiles failed to load because they return too much data.
+                    -- Alternative is to simplify polygons, but that takes more work/tuning.
+                    w.approx_area_sq_km >= (
+                        CASE
+                        WHEN z <= 5 THEN 100
+                        WHEN z <= 7 THEN 10
+                        WHEN z <= 8 THEN 5
+                        ELSE 0
+                        END
+                    )
                  GROUP BY w.geom,
                           w.pws_id,
                           w.pws_name
--- a/cdk/src/open-data-platform/data-plane/schema/schema.sql
+++ b/cdk/src/open-data-platform/data-plane/schema/schema.sql
@@ -46,7 +46,7 @@ CREATE OR REPLACE FUNCTION set_approx_area_sqkm_to_area_of_geom()
     RETURNS TRIGGER AS
 $$
 BEGIN
-    NEW.approx_area_sq_km = ST_Area(ST_Transform(new.geom, 3857)) / 1000000;
+    NEW.approx_area_sq_km = ST_Area(ST_Transform(new.geom, 4326)) / 1000000;
     RETURN NEW;
 END;
 $$ LANGUAGE plpgsql;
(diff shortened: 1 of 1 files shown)


kailamjeter (approved):

cdk/src/open-data-platform/data-plane/schema/schema.sql
```diff
@@ -41,6 +41,15 @@ BEGIN
 END;
 $$ LANGUAGE plpgsql;
 
+
+CREATE OR REPLACE FUNCTION set_approx_area_sqkm_to_area_of_geom()
+    RETURNS TRIGGER AS
+$$
+BEGIN
+    NEW.approx_area_sq_km = ST_Area(ST_Transform(new.geom, 3857)) / 1000000;
```
beekley, line 49: Why is this a different SRID than the 4326 used elsewhere?

beekley (approved):

cdk/src/open-data-platform/data-plane/schema/schema.sql
RP-3: 😱 ! This should _absolutely_ be 4326. Thanks!  was running on autopilot.

cdk/src/open-data-platform/data-plane/schema/schema.sql
```diff
@@ -428,7 +444,19 @@ BEGIN
                         SUM(w.service_connections_count)       AS service_connections_count,
                         SUM(w.population_served)               AS population_served
                  FROM water_systems w
-                 WHERE ST_Transform(w.bbox, 3857) && ST_TileEnvelope(z, x, y)
+                 WHERE
+                    ST_Transform(w.bbox, 3857) && ST_TileEnvelope(z, x, y) AND
```
beekley, line 448: Same here re: SRID.

RP-3: I think this one is legitimate.  `ST_TileEnvelope` always returns geometry in SRID 3857 (according to https://postgis.net/docs/ST_TileEnvelope.html), so IMO we should project to the same SRID for comparison. 

Does that seem fair?

beekley: SGTM. Can you add a comment with that info?

RP-3: Done. We actually legitimately use both 3857 and 4326 in several places, so I've added a block comment at the top explaining why and when we should use each one.

PR #141, opened 2022-08-29
Use BC's GA account ID

beekley:
## Description

Addresses: [Switch GA to use BC's account](https://tables.area120.google.com/table/bX_kMNBW8LA35n6l5Hz_co/row/aXx97ZAfHxzdYvhNHytkk6)

This switches the Google Analytics ID to [BC's account](https://lslr.slack.com/archives/C03D8ECRC3S/p1661276665032239?thread_ts=1660938852.883329&cid=C03D8ECRC3S).

### Changed

- GA ID.

## Testing and Reviewing

My [sandbox](https://beekley.leadout-sandbox.blueconduit.com/) didn't yell at my when I deployed this. But I don't have access to the GA dashboard to test that data is flowing.

### DIFF
--- a/client/public/index.html
+++ b/client/public/index.html
@@ -3,7 +3,7 @@
   <head>
     <!-- Conditionally include Google Analytics snippet, if targetting production -->
     <% if (process.env.NODE_ENV === 'production') { %>
-    <script async src="https://www.googletagmanager.com/gtag/js?id=UA-149759341-1"></script>
+    <script async src="https://www.googletagmanager.com/gtag/js?id=UA-146085874-1"></script>
     <script>
       window.dataLayer = window.dataLayer || [];
       function gtag() {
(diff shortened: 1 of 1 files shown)


client/public/index.html
```diff
@@ -3,15 +3,15 @@
   <head>
     <!-- Conditionally include Google Analytics snippet, if targetting production -->
     <% if (process.env.NODE_ENV === 'production') { %>
-    <script async src="https://www.googletagmanager.com/gtag/js?id=G-Y3Y8TPP27F"></script>
+    <script async src="https://www.googletagmanager.com/gtag/js?id=UA-149759341-1"></script>
```
jbhammon, line 6: When I open our GA dashboard, I only see `UA-146085874-1` as a tracking ID, not the one here. Am I looking in the wrong place, or not have access to the right GA view?

jbhammon (commented):
Changes make sense, but just had a question about the ID we're changing to.

client/public/index.html
beekley: I had copied it from other BC repos since I don't think I have access to that dashboard. But I see that Dunrie also used that ID so I made the switch.

jbhammon (approved):
New GA id seems fine to me.

PR #143, opened 2022-08-29
Alert on pipeline failure.

beekley:
## Description

Addresses: [Dev has monitoring](https://tables.area120.google.com/table/bX_kMNBW8LA35n6l5Hz_co/row/bYcpg5d5BPEdlvjUIGh_gZ)

Currently, devs have to manually check if a deployment succeeded. This PR adds a notification if the deployment pipeline failed.

I considered notifying on any changes, but that's noisy. I want to keep the signal:noise ratio high for the notifications Slack channels, which means they should only have messages if there is a problem. If it'd be useful to also notify on success, we can have another channel for that.

### New

- Notification on pipeline failure

## Testing and Reviewing

This successfully compiles via `npm run pipeline-synth`. But we won't know if it works until it's merged and deployed.

### DIFF
--- a/cdk/src/open-data-platform/app-plane/api/api-monitoring.ts
+++ b/cdk/src/open-data-platform/app-plane/api/api-monitoring.ts
@@ -1,0 +1,78 @@
+import { Construct } from 'constructs';
+import { RestApi } from 'aws-cdk-lib/aws-apigateway';
+import { Duration } from 'aws-cdk-lib';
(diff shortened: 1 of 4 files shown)


kailamjeter (approved):
Approving for pipeline changes but have an open question on the API change which also appears here.

kailamjeter:
Didn't realize this would automatically submit - going to leave this submitted and fine with making additional changes on API if needed separately.

PR #144, opened 2022-08-29
Add monitoring for API.

beekley:
## Description

Addresses: [Dev has monitoring](https://tables.area120.google.com/table/bX_kMNBW8LA35n6l5Hz_co/row/bYcpg5d5BPEdlvjUIGh_gZ) (based on the [Production Design](https://docs.google.com/document/d/1zZxCoXx5JzLXTOGVdvC4s8H-r82DFjfmNU-dmFPAQVI/edit#heading=h.cj0jtq6gzcgp))

This PR adds metrics and alarms the error rate for the entire API gateway. It's a bit of a naive implementation, since it smashes together all endpoints and all error codes. But I like that simplicity for now-- we can split it up into granular metrics in the future as we find it useful.

This defines alarms for "slow burn" (low error rate, but long duration) and "fast burn" (high error rate over a short duration). That way we can get an early warning of problems, but also see long-term issues too.

I have a feeling this is going to be noisy. But that might be a good thing. We can tune it as necessary.

### New

- Metrics and alarms for the API gateway.

## Testing and Reviewing

The alarms show up ([example](https://us-east-2.console.aws.amazon.com/cloudwatch/home?region=us-east-2#alarmsV2:alarm/beekley-OpenDataPlatformAppPlane-ApiStackAvailabilitySLOSlowBurn70D70F36-1NMZZJY6RA9A3?)) in my sandbox.

### DIFF
--- a/cdk/src/open-data-platform/app-plane/api/api-monitoring.ts
+++ b/cdk/src/open-data-platform/app-plane/api/api-monitoring.ts
@@ -1,0 +1,78 @@
+import { Construct } from 'constructs';
+import { RestApi } from 'aws-cdk-lib/aws-apigateway';
+import { Duration } from 'aws-cdk-lib';
+import { ITopic } from 'aws-cdk-lib/aws-sns';
+import * as cloudwatch from 'aws-cdk-lib/aws-cloudwatch';
+import { SnsAction } from 'aws-cdk-lib/aws-cloudwatch-actions';
+
+interface ApiMonitoringProps {
+  gateway: RestApi;
+  ticketSNSTopic?: ITopic;
+}
+
+export class ApiMonitoring extends Construct {
+  constructor(scope: Construct, id: string, props: ApiMonitoringProps) {
+    super(scope, id);
+
+    const { gateway, ticketSNSTopic } = props;
+
+    // Client errors are likely caused by problems in our client code since the API is not public,
+    // so they should be considered like an internal error.
+    // TODO: Split this into client and server errors once the API is pubished.
+    // TODO: Track the error budget, rather than looking at the error rate at any given time.
+
+    const availabilitySlo = 0.95;
+    const latencyMsSlo = 2000; // Miliseconds
+    const slowBurnPeriod = Duration.days(1);
+    const fastBurnPeriod = Duration.hours(1);
+
+    const alarms = [
+      new cloudwatch.MathExpression({
+        expression: '(clientError + serverError) / count',
+        label: 'Error Fraction',
+        period: slowBurnPeriod,
+        usingMetrics: {
+          clientError: gateway.metricClientError(),
+          serverError: gateway.metricServerError(),
+          count: gateway.metricCount(),
+        },
+      }).createAlarm(scope, 'AvailabilitySLOSlowBurn', {
+        alarmDescription: `The API has been running below its ${
+          availabilitySlo * 100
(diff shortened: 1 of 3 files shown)


cdk/src/open-data-platform/app-plane/api/api-monitoring.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    const { gateway, ticketSNSTopic } = props;
+
+    // Client errors are likely caused by problems in our client code since the API is not public,
+    // so they should be considered like an internal error.
+    // TODO: Split this into client and server errors once the API is pubished.
+    // TODO: Track the error budget, rather than looking at the error rate at any given time.
+
+    const availabilitySlo = 0.95;
+    const latencyMsSlo = 2000; // Miliseconds
```
kailamjeter, line 25: nit typo Milliseconds

cdk/src/open-data-platform/app-plane/api/api-monitoring.ts
```diff
@@ -0,0 +1,78 @@
...
+    const fastBurnPeriod = Duration.hours(1);
+
+    const alarms = [
+      new cloudwatch.MathExpression({
+        expression: '(clientError + serverError) / count',
+        label: 'Error Fraction',
+        period: slowBurnPeriod,
+        usingMetrics: {
+          clientError: gateway.metricClientError(),
+          serverError: gateway.metricServerError(),
```
kailamjeter, line 36: would it be worth having a separate alert for client vs server errors? I agree that alerting here for client errors makes sense for now, but it seems like these would probably be different issues so having both add up to 5% of total metrics might not be as significant of having 5% be just server or client errors (seems more likely the latter would signify a larger issue)

cdk/src/open-data-platform/app-plane/api/api-monitoring.ts
```diff
@@ -0,0 +1,78 @@
...
+          count: gateway.metricCount(),
+        },
+      }).createAlarm(scope, 'AvailabilitySLOSlowBurn', {
+        alarmDescription: `The API has been running below its ${
+          availabilitySlo * 100
+        }% SLO for ${slowBurnPeriod.toString()}.`,
+        evaluationPeriods: 1,
+        threshold: 1 - availabilitySlo,
+        treatMissingData: cloudwatch.TreatMissingData.NOT_BREACHING,
```
justinebreuch, line 45: Could you add a comment about what this is?

justinebreuch (approved):

cdk/src/open-data-platform/app-plane/api/api-monitoring.ts
beekley: I need spell check in vscode. Done.

cdk/src/open-data-platform/app-plane/api/api-monitoring.ts
beekley: I started to do so and realized we can drop this and use the default.

cdk/src/open-data-platform/app-plane/api/api-monitoring.ts
beekley: Yeah I think that's a good idea. As-is, this will be more sensitive than separating it out. I want to see how much this alarm goes off then we can tune (and possibly separate it). I have a TODO for that.

kailamjeter (approved):

PR #146, opened 2022-08-30
Fix parcels config to use proper property to read lead + fix zooming

justinebreuch:
## Description

Addresses: Add toggle between zip / water system geometry

Update parcel config to read public_lead_connections_low_estimate instead of old property
Update geolocate lambda to correctly convert bounding box --> min/max lat,long. Before it was swapping lat and lng, but it worked out because the FE also was. 

Zoom tight into parcels when an address is searched and found. 

## Testing and Reviewing

image

### DIFF
--- a/client/src/api/api_client.ts
+++ b/client/src/api/api_client.ts
@@ -51,8 +51,18 @@ class ApiClient {
   getGeoIds = async (lat: string, long: string): Promise<ApiResponse> => {
     return this.request(`${ApiClient.API_URL}/geolocate/${lat},${long}`, (data) => {
       return {
-        pwsId: data?.data?.water_system_pws_id,
-        zipCode: data?.data?.zip_code,
+        address: {
+          id: data?.data?.address.id,
+          boundingBox: data?.data?.address.bounding_box,
+        },
+        pwsId: {
+          id: data?.data?.water_system_pws_id?.id,
+          boundingBox: data?.data?.water_system_pws_id?.bounding_box,
+        },
+        zipCode: {
+          id: data?.data?.zip_code?.id,
+          boundingBox: data?.data?.zip_code?.bounding_box,
+        },
       };
     });
   };
--- a/client/src/components/NationwideMap.vue
+++ b/client/src/components/NationwideMap.vue
@@ -6,7 +6,7 @@
 </template>
 
 <script lang='ts'>
-import mapboxgl from 'mapbox-gl';
+import mapboxgl, { LngLatBounds } from 'mapbox-gl';
 import mapbox, { LngLatLike, MapLayerMouseEvent } from 'mapbox-gl';
 import MapLegend from './MapLegend.vue';
 import MapPopupContent from './MapPopupContent.vue';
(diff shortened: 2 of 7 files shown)


client/src/components/NationwideMap.vue
```diff
@@ -106,23 +107,37 @@ export default defineComponent({
...
-          lat: parseInt(this.geoState?.geoids?.lat),
-        };
+      const addressBoundingBox = this.geoState?.geoids?.address?.boundingBox;
+      const waterSystemBoundingBox = this.geoState?.geoids?.address?.boundingBox;
+
+      const lat = this.geoState?.geoids?.lat;
+      const long = this.geoState?.geoids?.long;
+
+      // If there's an address to zoom to, choose that.
+      if (addressBoundingBox != null && lat != null && long != null) {
```
justinebreuch, line 127: I don't yet want to use the bounding box here bc it zooms way too tight (all you see if the parcel). However, I think this will be useful to have

kailamjeter: Makes sense!

client/src/components/NationwideMap.vue
```diff
@@ -106,23 +107,37 @@ export default defineComponent({
...
-          [minLat, minLon],
-          [maxLat, maxLon],
-        ]);
-      } else if (this.geoState?.geoids?.lat != null && this.geoState?.geoids.long != null) {
-        const lonLat: LngLatLike = {
-          lon: parseInt(this.geoState?.geoids?.long),
-          lat: parseInt(this.geoState?.geoids?.lat),
-        };
+      const addressBoundingBox = this.geoState?.geoids?.address?.boundingBox;
+      const waterSystemBoundingBox = this.geoState?.geoids?.address?.boundingBox;
```
kailamjeter, line 121: pwsId

client/src/api/api_client.ts
```diff
@@ -51,8 +51,18 @@ class ApiClient {
   getGeoIds = async (lat: string, long: string): Promise<ApiResponse> => {
     return this.request(`${ApiClient.API_URL}/geolocate/${lat},${long}`, (data) => {
       return {
-        pwsId: data?.data?.water_system_pws_id,
-        zipCode: data?.data?.zip_code,
+        address: {
```
kailamjeter, line 54: right now if there is an address search we trigger getParcel if the geotype is address, but we could use this and only do the parcel search if there is an address present: https://github.com/BlueConduit/open-data-platform/blob/main/client/src/components/PredictionPanel.vue#L133 

would save an extra call to the API always since we can avoid doing water system fetch when an address is present on geoids too

kailamjeter (approved):

client/src/components/NationwideMap.vue
justinebreuch: Good catch thanks!

client/src/api/api_client.ts
justinebreuch: Done!

PR #150, opened 2022-08-31
Remove some animations to speed up scorecard load

beekley:
## Description

Addresses: [Disable FlyTo animation](https://tables.area120.google.com/table/bX_kMNBW8LA35n6l5Hz_co/row/8GbO7vkPsbr4aq4rYP6kbK)

Currently, the scorecard loads a nationwide map for a moment, which kicks off a bunch of irrelevant tile requests that are ultimately canceled by the client.

(Example: there are 65 tile requests and only 12 are for the parcel data that is displayed. Since this is a timing thing, the total number of requests is variable here and is sometimes lower)
 
image

This PR does a few main things:
1. Uses `jumpTo` instead of `flyTo` in the scorecard, to skip animating between addresses.
2. Initializes the map zoomed in to the global state's geoid, if present. So now it only starts tiles requests around the requested lat long.

(Example: there are only 12 tile requests and they're all displayed on the map)

image

As a result, the time to load a Toledo parcel scorecard drops from ~10s to ~3 s (without any cache!).

(an address without parcel data still only loads the relevant tiles. There's just a quick pop back out from the initial parcel-level zoom.)

## Testing and Reviewing

My sandbox doesn't have all of the parcels (oops). So I used Kaila's for the data during dev. That said, I spot checked a few places in my sandbox:

* [Toledo parcel](https://beekley.leadout-sandbox.blueconduit.com/scorecard/address/41.7176,-83.477155).
* [Address without parcel-level data](https://beekley.leadout-sandbox.blueconduit.com/scorecard/address/40.4581,-80.01223). (not sure why this still animates though)
* [Nationwide map](https://beekley.leadout-sandbox.blueconduit.com/map) still has animations and starts at default zoom and center.

### DIFF
--- a/client/src/components/NationwideMap.vue
+++ b/client/src/components/NationwideMap.vue
@@ -133,6 +133,7 @@ export default defineComponent({
 
       // If there's an address to zoom to, choose that.
       if (addressBoundingBox != null) {
+        // TODO: figure out why this sometimes still animates.
         if (this.scorecard) this.map?.jumpTo({ center, zoom: PARCEL_ZOOM_LEVEL });
         else this.map?.flyTo({ center, zoom: PARCEL_ZOOM_LEVEL });
 
@@ -353,6 +354,7 @@ export default defineComponent({
      */
     async createMap(): Promise<void> {
       // Start zoomed in for a scorecard to avoid unnecessary tile loads.
+      // TODO: pull the zoom level from the geoId in the global state.
       const zoom = this.scorecard ? PARCEL_ZOOM_LEVEL : DEFAULT_ZOOM_LEVEL;
       // The map is created before the state watcher fires, so get the center directly.
       const center = this.getLngLatFromState() ?? this.center;


client/src/components/NationwideMap.vue
```diff
@@ -336,55 +339,59 @@ export default defineComponent({
...
-
     },
 
     /**
      * Creates base Mapbox map and configures interactions and styles for
      * layers.
      */
     async createMap(): Promise<void> {
+      // Start zoomed in for a scorecard to avoid unnecessary tile loads.
+      const zoom = this.scorecard ? PARCEL_ZOOM_LEVEL : DEFAULT_ZOOM_LEVEL;
```
kailamjeter, line 356: nit: will this zoom out if we don't have parcel info? i.e. if we return just water system info, will this zoom out to zip level? if so we should make that default since that will be most cases.

kailamjeter (approved):

client/src/components/NationwideMap.vue
beekley: I didn't see a good ZIP zoom level to use, since the zoom depends on the size of the shape that the latlong points to. I added some TODOs to further optimize this.

PR #151, opened 2022-09-01
Geolocate fix

kailamjeter:
## Description

Addresses: Update scorecard page URL with geolocated address

Put lat/long in URL when user geolocates. This change affects both scorecard and map views.

## Testing and Reviewing

tested locally using geolocate functionality and going back and forth between that and the regular geocoder component.

### DIFF
--- a/client/src/components/NationwideMap.vue
+++ b/client/src/components/NationwideMap.vue
@@ -293,6 +293,9 @@ export default defineComponent({
         const long = result?.coords?.longitude;
         const baseUrl = this.scorecard ? SCORECARD_BASE : MAP_ROUTE_BASE;
 
+        // Ignore null lat/long.
+        if (lat == null || long == null) return;
+
         this.$router.push(`${baseUrl}/${GeoType.address}/${lat},${long}`);
       });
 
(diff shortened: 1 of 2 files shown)


client/src/components/NationwideMap.vue
```diff
@@ -267,18 +271,26 @@ export default defineComponent({
...
+        // Receive updates to the device's location as it changes.
+        trackUserLocation: true,
+      });
+
+      geolocateControl.on('geolocate', (result: any) => {
+        const lat = result?.coords?.latitude;
+        const long = result?.coords?.longitude;
+        const baseUrl = this.scorecard ? SCORECARD_BASE : MAP_ROUTE_BASE;
+
+        this.$router.push(`${baseUrl}/${GeoType.address}/${lat},${long}`);
```
beekley, line 289: Can `lat` or `long` be undefined here? (I don't really understand the `?` notation in ts yet) If so, let's make sure they have a value before pushing `undefined` to the URL.

beekley (approved):

client/src/components/NationwideMap.vue
kailamjeter: Done!

PR #153, opened 2022-09-06
Make the landing page responsive via a layout-only CSS library

beekley:
## Description

Addresses: [Mobile Optimization](https://tables.area120.google.com/table/bX_kMNBW8LA35n6l5Hz_co/row/9ZLTEnkXfJeetgENepdkVs)

(well, the [library itself](https://bulma.io/documentation/overview/) is more than just layout, but it can be [imported modularly](https://bulma.io/documentation/overview/modular/) so we only use the layout stuff.)

Previously, we set a minimum width so the page would at least render legibly on mobile. However, this meant the UI was quite small and required side-scrolling on narrow displays. 

This PR adds Bulma, which is a CSS library I've used before, to make the UI responsive. I started writing CSS to do this, but it got complicated quickly. So I'd rather just import some CSS that others have written to do it in a more robust way than I could.

Future work:

* Match Copper's spacing/etc by [overriding Bulma's variables](https://bulma.io/documentation/customize/with-node-sass/).
* Make the other pages responsive.

### New

- Added a SASS library, Bulma.

### Changed

- Pretty much every landing page component uses Bulma for layout.

## Testing and Reviewing

You can play around with the UI on my sandbox: https://beekley.leadout-sandbox.blueconduit.com/

image

image


client/src/components/ResponsiveNav.vue
```diff
@@ -0,0 +1,82 @@
...
+      [Titles.MAP_TITLE, router.MAP_ROUTE_BASE],
+      [Titles.ABOUT_TITLE, router.ABOUT_ROUTE],
+      [Titles.RESOURCES_TITLE, router.RESOURCES_ROUTE],
+    ];
+    return {
+      routes,
+    };
+  },
+  mounted() {
+    // Expand the menu via the burger button
```
kailamjeter, line 43: so this changes all navbars, not just mobile version?

client/src/components/landing_page/ExploreMapSection.vue
```diff
@@ -1,26 +1,38 @@
 <template>
-  <div class='section'>
-    <div class='header-section'>
-      <div class='h1-header-large'>
-        {{ messages.EXPLORE_MAP_SECTION_HEADER }}
+  <div class='section is-small'>
+    <div class='columns is-centered has-text-centered'>
```
kailamjeter, line 3: Not familiar with Bulma but it looks like these properties specify a lot of the properties we are currently specifying with CSS .. if this is the case can we remove duplicate css properties? (mixins for column, center, and text-align center all seem covered here)

client/src/components/ResponsiveNav.vue
beekley: Short answer: yes

Longer answer: there is only one navbar. It just looks different in mobile vs desktop (desktop shows the links while mobile shows an expandable menu via a burger button). Bulma doesn't include any JS so we have to write our own to make the button work here.

client/src/components/landing_page/ExploreMapSection.vue
beekley: Yup! Done. Good idea.

kailamjeter (approved):

client/src/components/ResponsiveNav.vue
kailamjeter: SG! Just wanted to make sure links still showed on desktop.

PR #155, opened 2022-09-06
Side panel

kailamjeter:
## Description

Addresses: 

- Add side panel from [mocks](https://www.figma.com/file/0LGxjZnPD3N0dCqHv8LRr6/NW-Map-Design-System-WIP?node-id=850%3A5844)
- Add Parcel Level View Separately from Zip/Water System Scenario

Adds the side panel from the mocks which shows just parcel info where available and zip / water system info where it is available and parcel info is not (as discussed with Nicole).

This change makes the existing loading bug very apparent so will probably look into implementing that next.

### New

- Side panel

### Changed

- Moved some shared logic around into util classes

### Removed

- N/A
## Testing and Reviewing

Deployed + Ran locally with addresses that have different levels of data including 
- [no WS data](https://kailajeter.leadout-sandbox.blueconduit.com/scorecard/address/44.189337,-68.66066)
- [just WS data](https://kailajeter.leadout-sandbox.blueconduit.com/scorecard/postcode/39.844983,-84.123508)
- [parcel data](https://kailajeter.leadout-sandbox.blueconduit.com/scorecard/address/41.716975,-83.47713)


client/src/components/SidePanel.vue
```diff
@@ -0,0 +1,136 @@
...
+      options: [] as ZoomLevel[],
+      selectedOption: null as ZoomLevel | null,
+      ZoomLevel,
+    };
+  },
+  methods: {
+    /**
+     * Returns the geo ID value corresponding to the given zoom level.
+     */
+    getGeoIdForZoomLevel(level: ZoomLevel): string | undefined {
```
beekley, line 54: I think it'd be good to format the name here in a more human-friendly way. For addresses, that might mean using title case. For water systems, can we print the water system name instead of its ID?

client/src/components/GeoIdSection.vue
```diff
@@ -0,0 +1,57 @@
+<template>
+  <div class='selected' v-if='selected'>
```
beekley, line 2: You can [bind class to a variable](https://vuejs.org/guide/essentials/class-and-style.html#binding-html-classes) instead of having separate divs for this.

beekley (approved):

client/src/components/SidePanel.vue
kailamjeter: Makes sense - put a TODO for water system because right now we aren't returning the name so this will take an API change (we also need to make a change there to prefer a water system with results, so I'll make both changes at once). 

For parcels I added a function to format the address in all lowercase except for the state USPS code. Additionally added text transform on the div to make it sentence case -- I think this combo is the least complex of the solutions. Ideally all transformation would happen in the function but manually capitalizing each word would be more complex than the css I think - let me know if you prefer this though.

client/src/components/GeoIdSection.vue
kailamjeter: thanks!

PR #156, opened 2022-09-07
Fix lead connections == 0 bug.

kailamjeter:
Fix bug introduced in https://github.com/BlueConduit/open-data-platform/pull/155/files# which removed the likelihood header when the likelihood == 0 bug.

Bug is currently visible at https://kailajeter.leadout-sandbox.blueconduit.com/scorecard/address/41.716255,-83.643702 but in the process of deploying the fix.

### DIFF
--- a/client/src/util/lead_data_util.ts
+++ b/client/src/util/lead_data_util.ts
@@ -30,8 +30,8 @@ export class LeadDataUtil {
    * @param prediction percent lead prediction
    */
   static formatPredictionAsLikelihood(prediction: number | null | undefined): string | null {
-    if (!prediction) return null;
-    
+    if (prediction == null) return null;
+
     switch (true) {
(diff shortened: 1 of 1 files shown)


client/src/util/lead_data_util.ts
```diff
@@ -30,8 +30,8 @@ export class LeadDataUtil {
    * @param prediction percent lead prediction
    */
   static formatPredictionAsLikelihood(prediction: number | null | undefined): string | null {
-    if (!prediction) return null;
-    
+    if (prediction == null) return null;
```
beekley, line 33: Should this check for `undefined` too? Since that's a possible value according to `prediction`'s type.

beekley (approved):

client/src/util/lead_data_util.ts
kailamjeter: Checking for null checks both! Intellij informed me it was redundant.

PR #159, opened 2022-09-08
API Fixes + Scorecard loading screen

kailamjeter:
## Description

Addresses: 

- Add loading indicator while the page loads data (both map and status): uses https://www.npmjs.com/package/vue-loading-overlay
- Modify API to prefer row with data (specifically water_systems, has multiple entries for some lat/longs)
- Bug with landing page Bulma implementation

Loading indicator:
This PR distinguishes water system data load status from parcel data load status in order to show the loading spinner before both sets of data return. This was causing water system data to render before parcel data in cases where there were both. 

Modify API to prefer row with data:
This sorts by # of service lines to prefer the intersecting water system which serves the largest population of people.

### Changed

- Single status in leadState to water system / parcel statuses

## Testing and Reviewing

tested locally + in sandbox: https://kailajeter.leadout-sandbox.blueconduit.com/

### DIFF
--- a/api/src/geolocate/get.handler.ts
+++ b/api/src/geolocate/get.handler.ts
@@ -5,6 +5,7 @@ import { ExecuteStatementRequest, SqlParametersList } from 'aws-sdk/clients/rdsd
 import { CORS_HEADERS } from '../util';
 
 const SCHEMA = 'public';
+const ID_FIELD = 'id';
 
 /**
  * Returns all geo identifiers that intersect with a lat, long.
@@ -20,7 +21,7 @@ async function getGeoDataForLatLong(
   params: SqlParametersList,
   geoid: string,
   table: string,
-  orderByField?: string,
+  orderByField: string = ID_FIELD,
   orderByAsc: boolean = true,
 ): Promise<BoundedGeoDatum | undefined> {
   // TODO(breuch): Consider updating this to all geoids when we support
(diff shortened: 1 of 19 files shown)


api/src/geolocate/get.handler.ts
```diff
@@ -5,19 +5,24 @@ import { ExecuteStatementRequest, SqlParametersList } from 'aws-sdk/clients/rdsd
...
+ * @param orderByField: field to sort results by.
+ * @param orderByAsc: whether to sort results in ascending order.
  */
 async function getGeoDataForLatLong(
   rdsService: RDSDataService,
   params: SqlParametersList,
   geoid: string,
   table: string,
+  orderByField: string = ID_FIELD,
+  orderByAsc: boolean = true,
```
justinebreuch, line 25: Does this ever resolve to true?

kailamjeter: We want this true by default I think - just orders IDs by ascending order. I don't think it actually matters if we are just ordering by ID but this makes it so that the non water system selectors aren't changed at all.

client/src/model/states/lead_data_state.ts
```diff
@@ -6,5 +6,6 @@ import { StatusState } from '@/model/states/status_state';
  */
 export interface LeadDataState {
   data?: LeadData;
-  status?: StatusState;
+  waterSystemStatus?: StatusState;
```
justinebreuch, line 9: Can you walk me through why we have both? Do we want to have separate errors or be in a different state when 1 fails vs both?

kailamjeter: For parcels we request both sets of data (water system + parcel level) and they don't return to the FE at the same time, so if we have just one status it will be made success from pending when the first api call returns (at least the way its currently set up, not sure how we would modify a single status to keep track of both calls with this pattern). We want to wait until both return to show any data though, since we will show one or the other depending on whether parcel data is present. Let me know if this doesn't make sense and we can chat over video.

justinebreuch: Can you leave a TODO to create an error when no rows are present and then determine how to have 2 unified status for lead data that is only set to success when both values came through?

justinebreuch (approved):

client/src/model/states/lead_data_state.ts
kailamjeter: Done!

PR #160, opened 2022-09-09
Add support for New Orleans + Richmond to LSLR section

kailamjeter:
## Description

Addresses: CTA#5: Display a button to sign up for an LSL replacement in pilot cities (toledo, richmond).

- Add bounding boxes for Richmond + New Orleans to intersect the lat,long with for surfacing lslr in supported cities.
- Capitalize Toledo in parcels import. (Will update dev DB when this is submitted)

### New

- Bounding boxes + getCityForLatLong in geo data util.

### Changed

- showLslrSection logic (From watcher + data variable to computed property)

## Testing and Reviewing

Tested locally in all 3 cities, both addresses and zip codes. Tested outside of those cities too to make sure nothing shows up when there is no result.

### DIFF
--- a/client/src/util/geo_data_util.ts
+++ b/client/src/util/geo_data_util.ts
@@ -1,5 +1,22 @@
-import { GeoData } from '@/model/states/model/geo_data';
+import { City, GeoData } from '@/model/states/model/geo_data';
 import { ZoomLevel } from '@/model/states/model/map_data';
+import { LngLatBounds, LngLatLike } from 'mapbox-gl';
+
+// Define Toledo geometry bounding box to restrict parcel data layer to Toledo.
+// This is needed because we only have parcel-level predictions for Toledo, so this data layer will
+// be empty outside of these boundaries.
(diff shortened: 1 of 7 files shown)


client/src/util/geo_data_util.ts
```diff
@@ -36,4 +52,26 @@ export class GeoDataUtil {
     }
     return options;
   }
+
+  /**
+   * Gives the city which contains the given lat, long if it is a supported city
+   * (Richmond, New Orleans, Toledo).
+   */
+  static getCityForLatLong(latString: string | undefined, longString: string | undefined): City | null {
+    const lat = Number(latString);
```
justinebreuch, line 61: Curious if I should be using this instead of parseInt or float? Not blocking

justinebreuch (approved):

client/src/util/geo_data_util.ts
kailamjeter: I think they are pretty much the same!

PR #161, opened 2022-09-12
Allow for nulls in water systems and parcel tables

justinebreuch:
## Description

Allow for nulls in water system + parcel table. Before we just wrote 0 when the file had null, but this distinction is meaningful. 0 means confirmed value. Null means there is no prediction.

### Changed

Both tables now have  isNull fields set if the respective field in the row is undefined. This allows us to write nulls with the RDS service

## Testing and Reviewing

Dropped tables and re-imported


cdk/src/open-data-platform/data-plane/data-import/write-water-systems-data.handler.ts
```diff
@@ -85,25 +94,23 @@ function getTableRowFromRow(row: any): SqlParametersList | null {
...
-    // Otherwise, get estimate from prediction interval.
-  } else {
+  const leadCountIsNotReported = leadConnectionsCount == null || leadConnectionsCount < 0;
+  const lowEstimate = getValueOrDefault(properties.low);
+  const highEstimate = getValueOrDefault(properties.high);
+  if (leadCountIsNotReported && lowEstimate != null && highEstimate != null) {
     // TODO: Switch to storing low, high in db.
-    leadConnectionsCount =
-      (getValueOrDefault(properties.low) + getValueOrDefault(properties.high)) / 2;
+    leadConnectionsCount = (lowEstimate + highEstimate) / 2;
```
kailamjeter, line 105: since we are editing this now would it be a good time to differntiate between known / predicted lead connections in db? We could add an additional column to track each or just a column which tracks isConfirmed

kailamjeter (approved):

cdk/src/open-data-platform/data-plane/data-import/write-water-systems-data.handler.ts
```diff
@@ -85,25 +94,23 @@ function getTableRowFromRow(row: any): SqlParametersList | null {
...
   }
   return (
     new WaterSystemsTableRowBuilder()
       .pwsId(properties.pwsid)
       .pwsName(properties.pws_name ?? '')
       .leadConnectionsCount(leadConnectionsCount)
       .serviceConnectionsCount(getValueOrDefault(properties.service_connections_count))
-      .populationServed(getValueOrDefault(properties.population_served))
+      .populationServed(getValueOrDefault(properties.population_served) ?? 0)
```
beekley, line 113: Why do we want population to default to 0 if there's no data?

beekley (approved):
Make sure to re-import this data to dev once it has deployed.

cdk/src/open-data-platform/data-plane/data-import/write-water-systems-data.handler.ts
justinebreuch: Good catch. This value is already nullable so didn't need this ?? 0

cdk/src/open-data-platform/data-plane/data-import/write-water-systems-data.handler.ts
justinebreuch: Added an is_estimate column based on leadCountIsNotReported

PR #162, opened 2022-09-12
Remove popup

justinebreuch:
## Description

Does not create popup when on the scorecard page. 

## Testing and Reviewing

In UI


client/src/components/NationwideMap.vue
```diff
@@ -286,7 +286,7 @@ export default defineComponent({
      * Sets up interaction handlers for map.
      */
     setUpInteractionHandlers(): void {
-      if (this.map == null) return;
+      if (this.map == null || this.scorecard) return;
```
justinebreuch, line 289: I'd lke to rename this like "basicMap" instead of scorecard bc this component should not have to know about the pages it is served on

kailamjeter: that makes sense! do you want to do separately? Feel like should be small enough to do in this PR.

kailamjeter (approved):

beekley (approved):

client/src/components/NationwideMap.vue
justinebreuch: Updated!

PR #164, opened 2022-09-12
Call city implementation

kailamjeter:
## Description

Addresses: CTA#2: Display contact info to call the city or water system 

Also fixes bug introduced in https://github.com/BlueConduit/open-data-platform/pull/157 which looks at the parcel city field instead of the computed property for the lslr section. 

### New

- Contact your city section

## Testing and Reviewing

Tested locally and in sandbox: 

[Toledo](https://kailajeter.leadout-sandbox.blueconduit.com/scorecard/address/41.724295,-83.552465) 
[New Orleans](https://kailajeter.leadout-sandbox.blueconduit.com/scorecard/postcode/29.960244,-90.073635)
[Richmond](https://kailajeter.leadout-sandbox.blueconduit.com/scorecard/postcode/37.579738,-77.53986)

### DIFF
--- a/client/src/components/ContactCitySection.vue
+++ b/client/src/components/ContactCitySection.vue
@@ -22,29 +22,39 @@ import { City } from '../model/states/model/geo_data';
 import { defineComponent, PropType } from 'vue';
 import { ScorecardMessages } from '../assets/messages/scorecard_messages';
 
-interface CityInformation {
+interface CityContactInfo {
   name: City;
   website: string;
   phoneNumber: string;
 }
 
-const CITY_INFO: CityInformation[] = [
-  {
-    name: City.newOrleans,
-    website: 'https://www.swbno.org/DrinkingWater/LeadAwareness',
-    phoneNumber: '(504) 529-2837',
-  },
-  {
-    name: City.richmond,
-    website: 'https://www.vdh.virginia.gov/richmond-city/healthy-homes',
-    phoneNumber: '(804) 205-3726',
-  },
-  {
-    name: City.toledo,
-    website: 'https://toledo.oh.gov/residents/water/lead-service-lines',
-    phoneNumber: '(419)-936-2020',
-  },
-];
+// TODO: Move to table in DB if we get more entries.
+const CITY_INFO: Map<City, CityContactInfo> = new Map([
+  [
+    City.newOrleans,
+    {
+      name: City.newOrleans,
+      website: 'https://www.swbno.org/DrinkingWater/LeadAwareness',
+      phoneNumber: '(504) 529-2837',
+    },
+  ],
+  [
+    City.richmond,
+    {
+      name: City.richmond,
+      website: 'https://www.vdh.virginia.gov/richmond-city/healthy-homes',
+      phoneNumber: '(804) 205-3726',
+    },
+  ],
+  [
+    City.toledo,
+    {
+      name: City.toledo,
+      website: 'https://toledo.oh.gov/residents/water/lead-service-lines',
+      phoneNumber: '(419)-936-2020',
+    },
+  ],
+]);
 
 /**
  * Component for the contact city section of scorecard.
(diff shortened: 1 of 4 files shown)


client/src/components/ContactCitySection.vue
```diff
@@ -0,0 +1,89 @@
...
+import { defineComponent, PropType } from 'vue';
+import { ScorecardMessages } from '../assets/messages/scorecard_messages';
+
+interface CityInformation {
+  name: City;
+  website: string;
+  phoneNumber: string;
+}
+
+const CITY_INFO: CityInformation[] = [
```
beekley, line 31: Did you consider using a map here instead of a list (e.g. `Map<City, CityInformation>`)? So we don't have to iterate through the list to look it up. 

(This is a really a nit because iterating through a list of 3 objects is basically free).

client/src/components/ContactCitySection.vue
```diff
@@ -0,0 +1,89 @@
...
+  </div>
+</template>
+
+<script lang='ts'>
+
+import { City } from '../model/states/model/geo_data';
+import { defineComponent, PropType } from 'vue';
+import { ScorecardMessages } from '../assets/messages/scorecard_messages';
+
+interface CityInformation {
```
beekley, line 25: nit: I'd rather name this something more specific to its purpose, such as `ContactInfo`.

client/src/components/ContactCitySection.vue
beekley, line 31: Also, do you know if there's any intention of getting this kind of data for more cities? If so, let's add a TODO to move this to a DB somewhere once the list gets longer than a few elements.

client/src/components/NationwideMap.vue
```diff
@@ -99,7 +99,7 @@ export default defineComponent({
     center: {
       // There is no constructor function for a Tuple, so use object here and
       // cast to PropType of a tuple.
-      // See https://vuejs.org/guide/typescript/options-api.html#typing-component-props.
+      // See https://vuejsForg/guide/typescript/options-api.html#typing-component-props.
```
beekley, line 102: Is this change intentional?

client/src/views/ScorecardView.vue
```diff
@@ -42,6 +42,11 @@
         </div>
       </div>
 
+      <ContactCitySection
```
beekley, line 45: Is there a reason this isn't in the section above, which has the other CTAs?

beekley (approved):
Left a few minor questions/comments.

client/src/components/ContactCitySection.vue
kailamjeter: I did, but have been hating javascript maps so was trying to avoid it (the initialization of the map to me is just so excessive). I'll change it though it is just a better way to do it.

client/src/components/NationwideMap.vue
kailamjeter: oh no! didn't even realize I must have been command f'ing or something. Will revert!

client/src/views/ScorecardView.vue
kailamjeter: I tried putting it inside of that div as another 1/3 of the page element but it looked really strange: https://screenshot.googleplex.com/9Ft8xw6DsXcwL5J I think because visually it is different from those other ctas so it looks strange to me that they are grouped in a row like that (as opposed to https://screenshot.googleplex.com/6bsiMhQtph5iZzg). 

Let me know what you think!

client/src/components/ContactCitySection.vue
kailamjeter: Change to CityContactInfo

client/src/components/ContactCitySection.vue
kailamjeter: Not sure honestly, probably depends on user feedback and wouldn't happen before our fellowship is over. Left a TODO in case BC ends up adding more.

PR #166, opened 2022-09-12
Fix side panel width

beekley:
## Description

When making the scorecard responsive, the side panel got a dynamic-width, which looks weird on a side screen.

image

This PR fixes the width.

### Changed

- Side panel has a max-width.
- GeoIdSections have a dynamic width to fill the side panel. Consequently, they are now wider in the mobile view when not constrained next to the map.

## Testing and Reviewing

Locally tested at different screen sizes.

Mobile:

image

Tablet:

image

Desktop:

image

### DIFF
--- a/client/src/views/ScorecardView.vue
+++ b/client/src/views/ScorecardView.vue
@@ -10,11 +10,18 @@
       <!-- Only display the side-panel, full-width on mobile. -->
       <ScorecardMapSearchBar class='is-hidden-mobile' />
       <div class='columns is-variable is-centered is-gapless'>
-        <div class='column side-panel' v-if='showResults'>
+        <!-- Conditionally use the "side-panel" class to limit column width on desktop only -->
+        <div class='column is-hidden-desktop' v-if='showResults'>
           <div class='section'>
             <SidePanel />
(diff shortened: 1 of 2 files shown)


client/src/views/ScorecardView.vue
```diff
@@ -2,21 +2,20 @@
...
     </div>
     <div v-if='showScorecard'>
       <PredictionPanel />
 
       <!-- Only display the side-panel, full-width on mobile. -->
       <ScorecardMapSearchBar class='is-hidden-mobile' />
-      <div class='columns is-variable is-centered'>
-        <div class='column'>
+      <div class='columns is-variable is-centered is-gapless'>
+        <div class='column side-panel'>
```
kailamjeter, line 13: do you know if this fixes the issue where the side panel shows up when theres no result: https://screenshot.googleplex.com/8h6U8AZfjy8wDoN? If not could you do that change here as well?

beekley: Done.

kailamjeter (approved):

justinebreuch (approved):

PR #169, opened 2022-09-14
Display UI feedback when a user clicks the copy scorecard button

justinebreuch:
## Description

Addresses: Display UI feedback when a user clicks the copy scorecard button.

Adds feedback to copy to clipboard button with a tooltip. See screenshot. Disappears after 1 second. 

### New

Added popper for tooltips to appear to give feedback that copy to clipboard succeeded.

## Testing and Reviewing

image


client/src/components/ActionSection.vue
```diff
@@ -45,4 +72,25 @@ export default defineComponent({
 .explain-text {
   padding: $spacing-md;
 }
+
+.icon {
+  max-width: $spacing-md;
+}
+
+// These variables cannot reference css vars
+.tooltip-content {
```
beekley, line 81: Can you use the copper design tokens for any of this?

beekley (approved):

client/src/views/ScorecardView.vue
```diff
@@ -2,7 +2,8 @@
   <div>
     <!-- Cover the entire page with loading element until data is ready. -->
     <div class='loading' v-if='!showScorecard'>
-      <loading :active='true' :is-full-page='false' color='#2553A0' loader='bars' opacity='1' />
+      <loading :active='true' :is-full-page='false' color='#2553A0'
```
kailamjeter, line 5: can we put each property on a newline like 
<loading :active='true'
                :is-full-page='false'
                :color='' 
.... 
or put back to single line instead of splitting down the middle

kailamjeter (approved):

client/src/components/ActionSection.vue
justinebreuch: for some reason these css var properties don't work with css vars

client/src/views/ScorecardView.vue
justinebreuch: Done!

PR #172, opened 2022-09-14
Show percent lead lines instead of # lead service lines and # service lines in popup 

justinebreuch:
## Description

Addresses: Make lead service lines a percentage in popup

Show percent lead lines instead of # lead service lines and # service lines in popup 

### New

Computed fields as opposed to features directly from the DB

## Testing and Reviewing

image


client/src/data_layer_configs/lead_service_lines_by_water_systems_config.ts
```diff
@@ -114,22 +115,23 @@ const popupInfo: PopupInfo = {
...
     },
   ],
+  computedProperties: [
+    {
+      label: 'Percent lead lines',
+      name: 'percent_lead',
+      dataType: FeaturePropertyDataType.Percentage,
+      calculate: (features: GeoJsonProperties) =>
+        features == null
+          ? null
```
kailamjeter, line 131: should we check that this is not 0? would this give an error or just be 0

client/src/data_layer_configs/lead_service_lines_by_water_systems_config.ts
```diff
@@ -114,22 +115,23 @@ const popupInfo: PopupInfo = {
...
+          : features.lead_connections_count / features.service_connections_count,
```
kailamjeter, line 132: are these strings? does it cast to int when you divide?

kailamjeter (approved):

client/src/data_layer_configs/lead_service_lines_by_water_systems_config.ts
```diff
@@ -114,22 +115,23 @@ const popupInfo: PopupInfo = {
...
-    },
     {
       label: 'Population served',
       name: 'population_served',
       dataType: FeaturePropertyDataType.Number,
     },
   ],
+  computedProperties: [
+    {
+      label: 'Percent lead lines',
```
beekley, line 126: Let's make this `Estimated percent lead lines` per [this](https://tables.area120.google.com/table/bX_kMNBW8LA35n6l5Hz_co/row/aVsp8DFDsMieWlDzLIJk8H).

(though I still don't love this phrasing. But something explicit like  `Estimated fraction of water service lines that contain lead` feels too long. I'll see if someone from branding can look through and possibly edit the text throughout the app)

beekley (approved):

client/src/data_layer_configs/lead_service_lines_by_water_systems_config.ts
justinebreuch: Good note thanks

client/src/data_layer_configs/lead_service_lines_by_water_systems_config.ts
justinebreuch: I don't think so, when I look at them they are raw numbers

client/src/data_layer_configs/lead_service_lines_by_water_systems_config.ts
justinebreuch: Done!

PR #173, opened 2022-09-14
Small UI fixes

kailamjeter:
## Description

Addresses: 

- Searching on Population / Violations data layers takes you back to water systems layer when zooming to a result
- Bug where you have to click twice on data layer to see a layer change on nationwide map

## Testing and Reviewing

Tested locally + in sandbox: https://kailajeter.leadout-sandbox.blueconduit.com/

Clicked around nationwide map, switched layers when zoomed into address, searched new address in violations layer.

### DIFF
--- a/client/src/model/slices/map_data_slice.ts
+++ b/client/src/model/slices/map_data_slice.ts
@@ -42,7 +42,7 @@ const mapSlice = createSlice({
     },
     mapDataCleared: () => {
       return {
-        data: {},
+        mapData: { dataLayers: Array.from(ALL_DATA_LAYERS.keys()) },
         status: { status: Status.success },
       };
     },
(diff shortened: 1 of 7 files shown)


client/src/model/slices/map_data_slice.ts
```diff
@@ -42,7 +42,7 @@ const mapSlice = createSlice({
     },
     mapDataCleared: () => {
       return {
-        data: {},
+        mapData: { dataLayers: Array.from(ALL_DATA_LAYERS.keys()) },
```
justinebreuch, line 45: Good call

kailamjeter: Yeah was a sneaky bug...

justinebreuch (approved):

beekley:
(I'm merging this while Kaila is OOO)

kailamjeter:
Thanks for merging this!

PR #174, opened 2022-09-15
Add canaries that probe for basic functionality

beekley:
## Description

Addresses: [Create canaries (probes) that verify the frontend works in each environment](https://tables.area120.google.com/table/bX_kMNBW8LA35n6l5Hz_co/row/826xIGWGqdF62fj6Icp__4)

_Short, descriptive summary of the changes_

### New

- _Any new things_

### Changed

- _Any changed things_

### Removed

- _Any removed things_

## Testing and Reviewing

_Add descriptions, steps, or checklists for how someone can review this PR to verify it's working as intended_


cdk/src/monitoring/synthetics.handler.ts
```diff
@@ -0,0 +1,50 @@
...
+  // Navigate to the initial url
+  await synthetics.executeStep('navigateToUrl', async function (timeoutInMillis = 30000) {
+    await page.goto(url, { waitUntil: ['load', 'networkidle0'], timeout: timeoutInMillis });
+  });
+
+  // Fill and search for an address.
+  await synthetics.executeStep('searchAddress', async function () {
+    await page.waitForSelector("[id='app']", { timeout: 30000 });
+    await page.screenshot();
+    const searchInput = await page.waitForSelector('input.mapboxgl-ctrl-geocoder--input');
```
justinebreuch, line 31: nit: Can you move some of these to constants?

cdk/src/monitoring/synthetics.handler.ts
```diff
@@ -0,0 +1,50 @@
...
+    await searchInput.type(address);
+    const suggestion = await page.waitForSelector('.mapboxgl-ctrl-geocoder--suggestion-title');
+    await page.screenshot();
+    await suggestion.click();
+    // This selects the first button with this class.
+    // TODO: add and use element IDs.
+    const searchButton = await page.waitForSelector('#scorecard-search-button');
```
justinebreuch, line 38: Curious if this is the recommended way of identifying buttons (via ids) rather than their text? This would mean we have to add ids for the sole purpose of testing

justinebreuch (approved):

cdk/src/monitoring/synthetics.handler.ts
beekley: Good point, and I don't know if it's the best way of identifying elements. I did a quick Google and Moma search for best practices and found [this talk](https://docs.google.com/presentation/d/19sqLP_cuBPWK7ol5SrUaQL7_L0hpn1u0ewmODXFg42Q/edit#slide=id.p), which didn't seem to state a preference (it identified elements by class or id in the examples).

I'll leave it as-is and let devs decide how they want to in future tests.

cdk/src/monitoring/synthetics.handler.ts
beekley: I'd rather leave it inline because it's only referenced once, here.

PR #177, opened 2022-09-16
Add another GA snippet

beekley:
## Description

The existing snippet would require us devs to manually tag elements in the frontend for tracking. This one loads the frontend in a "container" so it can be tagged via the GA web UI without dev involvement.

### Testing

Deployed to [sandbox](https://beekley.leadout-sandbox.blueconduit.com/) and it didn't explode.


beekley:
Waiting for the new container ID

kailamjeter (approved):

PR #178, opened 2022-09-16
Re-Add canaries that probe for basic functionality

beekley:
(re-write of https://github.com/BlueConduit/open-data-platform/pull/174, which I reverted in https://github.com/BlueConduit/open-data-platform/pull/176. The fix was to use `synthetics.Schedule.cron` method, not `.expression`)

## Description

Addresses: [Create canaries (probes) that verify the frontend works in each environment](https://tables.area120.google.com/table/bX_kMNBW8LA35n6l5Hz_co/row/826xIGWGqdF62fj6Icp__4)

We've sometimes broken dev and were notified by users (i.e. other fellows). This PR sets up automation that checks that users can interact with the deployed application, through a headless browser running periodically in AWS.

[`cdk/src/monitoring/synthetics.handler.ts`](https://github.com/BlueConduit/open-data-platform/pull/174/files#diff-0ba1fb251661e1bfa2f24762b014b0554e39f05f8f74786caa04b1f1ae5fe48d) defines the test code. It uses [Puppeteer](https://pptr.dev/api/puppeteer.page) to interact with the page and assert stuff.

It runs periodically in non-sandbox environments (e.g. dev). But only once in sandbox since we only have one monitoring stack for all sandbox envs. If anyone wants to test it in their sandbox, they can either run `npm run monitoring-deploy` or go to the canary dashboard and edit the env variables to point to their sandbox.

image

### New

- `SyntheticsStack`, which deploys the canary (i.e. probe, i.e. "CloudWatch Synthetics")

### Changed

- Added an ID to a button element, so the canary can select it directly rather than by class name.
- Moved domain name stuff to util, so monitoring has access to it.

## Testing and Reviewing

You can see the canary pass in [CloudWatch](https://us-east-2.console.aws.amazon.com/cloudwatch/home?region=us-east-2#synthetics:canary/detail/monitoringsynth81c9cc).

This is based on a canary I manually wrote in the console for dev:
image


cdk/src/monitoring/synthetics.handler.ts
```diff
@@ -0,0 +1,49 @@
+const synthetics = require('Synthetics');
+const log = require('SyntheticsLogger');
+const syntheticsConfiguration = synthetics.getConfiguration();
+
+exports.handler = async function () {
+  // TODO: cycle through several addresses
```
kailamjeter, line 6: i would say priority for next tests would be to get coverage across the different types of results users might see, i.e. two more addresses, one with just water system data (no parcel) and one with neither parcel nor water system data.

cdk/src/monitoring/synthetics.handler.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  await synthetics.executeStep('viewScorecard', async function () {
+    // Check for prediction presence.
+    await page.waitForFunction(
+      'document.querySelector("body").innerText.includes("Low likelihood")',
+    );
+    // Check for side panel presence.
+    await page.waitForSelector('.geoid-section');
+    await page.screenshot();
+  });
```
kailamjeter, line 48: should we also check for the explaining your score section and CTAs?

cdk/src/monitoring/synthetics.handler.ts
beekley: Good idea. I've made this comment more verbose to add those cases.

cdk/src/monitoring/synthetics.handler.ts
beekley: Good idea. I added some basic presence tests.

client/src/components/landing_page/ScorecardSearch.vue
```diff
@@ -6,7 +6,12 @@
       :acceptedTypes='acceptedTypes'
       @result='onGeocodeResults'
     />
-    <button class='gold-button' @click='onSearch' :disabled='!searchHasResult'>
+    <button
+      class='gold-button'
+      id='scorecard-search-button'
```
justinebreuch, line 11: Is there a way to search this button by text rather than adding an id?

beekley: There probably is. But I didn't see an obvious way to select the button element with text. I tried using a `waitForFunction` like I use for finding text on the page, but the canary failed with:

> TypeError: searchButton.click is not a function

So I went with ID to get it working.

PR #180, opened 2022-09-19
Add unit test for About page

justinebreuch:
## Description

Addresses: Add unit test

Add [Vitest testing framework](https://vitest.dev/) to the client 

### New

vitest.config to setup testing settings for all unit tests inside client/src/
AboutUsViewTest - a basic test on our simplest page as a proof of concept / model to follow. Will try to add unit tests to more important pages. They originally were not working because the testing framework could not resolve dynamic images. Will need to look more deeply into this.

## Testing and Reviewing

Added to the README, but run npm test -watch to run / hot reload tests


kailamjeter (approved):
Nice! Thanks for this!

PR #182, opened 2022-09-20
Add email signup

kailamjeter:
## Description

Addresses: Orange/Red paths: show an email sign-up when there is no parcel-level data or no data at all for a given address.

Adds email signup to landing page and scorecard when there is no parcel data. 

Note that this does not include a thank you page which is needed to unblock releasing this to users. Currently it shows a message saying we need a welcome page. That will be done in a separate PR.

### New

- EmailSignup component and types declaration which extends the Window type to include the hbspt variable

## Testing and Reviewing

Tested locally and deployed to sandbox: https://kailajeter.leadout-sandbox.blueconduit.com/ tested addresses in Toledo (should not show up in scorecard) and elsewhere (should show up everywhere else as we do not have parcel data).

View [screenshot](https://screenshot.googleplex.com/3cyXH88T9Ed9ZVS).


client/src/components/EmailSignup.vue
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Section containing form to sign up for LeadOut data updates.
+ *
+ * This wraps a hubspot form which is configured and styled in the hubspot editor.
+ */
+export default defineComponent({
+  name: 'EmailSignup',
+  data() {
+    return {
+      MESSAGES: LandingPageMessages,
```
justinebreuch, line 28: nit: for consistency, camelcase the data vars

client/src/components/EmailSignup.vue
```diff
@@ -0,0 +1,77 @@
...
+    const formScript = document.createElement('script');
+    document.body.appendChild(script);
+    document.body.appendChild(formScript);
+
+    script.addEventListener('load', () => {
+      if (window.hbspt) {
+        // LeadOut Website Subscribe Form.
+        window.hbspt.forms.create({
+          region: 'na1',
+          portalId: '7810253',
```
justinebreuch, line 46: Do any of these need to be kept secret via ENV?

justinebreuch (approved):

client/src/components/EmailSignup.vue
kailamjeter: Not that I'm aware of - basically hubspot says copy / paste this to embed in your app, so I think they are all just related to the form, i.e. don't think that they reveal any private info about BCs account. I can double check with Dunrie though.

client/src/components/EmailSignup.vue
kailamjeter: done!

client/src/components/EmailSignup.vue
kailamjeter: Checked with Dunrie and she isn't sure, but doesn't seem like a concern. Resolving this.

PR #183, opened 2022-09-20
Thank you page finished.

kailamjeter:
## Description

Addresses: Adds thank you page to be routed to after users subscribe to LeadOut.

Right now there is no entrypoint to this page in the app - this will be hooked up to the hubspot form in the editor.

### New

- ThankYouView component.

## Testing and Reviewing

ran locally and routed to http://localhost:8080/subscribed 

[screenshot](https://screenshot.googleplex.com/9wKpnED7MoYbayi)


beekley (approved):
LGTM.

I feel there should be some action for the user to take after, even if it's just a link to go back home. That might be something for Product though.

PR #184, opened 2022-09-21
Switch income percentiles for scorecard summary

justinebreuch:
## Description

Addresses: Investigate whether low or high income index mean higher income (get clarity and update content for this)

Lower percentiles are lower income. Updated copy to reflect that.


beekley (approved):
Should we update anything the UI to reflect this change?

justinebreuch:
> 

This updates the labels so that the lower percentiles now say "Highly disadvantages" instead of "Less disadvantaged" so the UI now reflects the real intent of the API data

PR #185, opened 2022-09-21
Add tests for api client 

justinebreuch:
## Description

Addresses: Add unit test for API client

Adds checks to API client that it returns/converts proper fields from API gateway response. Will be using this mock library to test that the app state gets updated correctly based on the API responses (in a separate PR)

### New

1. api_client.test which has the tests themselves
2. test_data which stores mock API responses 

### Changed
Updated API with mistakes found by testing :)

## Testing and Reviewing

There's scripts to run the test. Can also use the UI testing library (running `vitest --ui`)


beekley (approved):
Nice work!

PR #190, opened 2022-09-22
Use common messages file in the canaries

beekley:
## Description

Addresses: [Canaries](https://tables.area120.google.com/table/bX_kMNBW8LA35n6l5Hz_co/row/826xIGWGqdF62fj6Icp__4)

The canaries started failing after https://github.com/BlueConduit/open-data-platform/pull/186 changed a string. Since the canary handler is a single JS blob, this PR does some string manipulation to combine the handler with the messages file so that the handler can reference the same `ScorecardMessages` as the client.

### Changed

- Semi-hacky transpiling method of importing the handler.
- Replaced strings in the handler with references to `ScorecardMessages` properties.

## Testing and Reviewing

The [canaries](https://us-east-2.console.aws.amazon.com/cloudwatch/home?region=us-east-2#synthetics:canary/detail/monitoringsynth81c9cc) pass in my sandbox!

image

### DIFF
--- a/cdk/src/monitoring/synthetics.handler.ts
+++ b/cdk/src/monitoring/synthetics.handler.ts
@@ -1,5 +1,11 @@
 const synthetics = require('Synthetics');
 const syntheticsConfiguration = synthetics.getConfiguration();
+const log = require('SyntheticsLogger');
+
+// This is implicitly imported in synthetics.ts. TS doesn't like that, so ignore the error that this
+// doesn't exist.
+// @ts-ignore
+const scorecardMessages = ScorecardMessages;
 
 /**
  * Helper function for searching the entire page for a string. Returns if the string was found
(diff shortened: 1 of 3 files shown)


cdk/src/monitoring/synthetics.handler.ts
```diff
@@ -54,14 +60,14 @@ exports.handler = async function () {
...
     // Check for side panel presence.
     await page.waitForSelector('.geoid-section');
     // Check for CTA presence.
-    await waitForString(page, 'Research water filters');
-    await waitForString(page, 'Copy scorecard link');
+    await waitForString(page, scorecardMessages.RESEARCH_WATER_FILTERS);
+    await waitForString(page, scorecardMessages.COPY_TO_CLIPBOARD);
     // Check for score details presence.
-    await waitForString(page, 'Understanding your score');
+    await waitForString(page, scorecardMessages.SCORECARD_SUMMARY_PANEL_HEADER(null));
```
justinebreuch, line 70: I think you actually do want to pass in a zip code here since it should say the zip code for Toledo

justinebreuch (approved):

cdk/src/monitoring/synthetics.handler.ts
beekley: Good point, done.

beekley: Actually, the test fails when I pass a ZIP here. Is that expected? Here's the scorecard for that address: https://beekley.leadout-sandbox.blueconduit.com/scorecard/address/41.717025,-83.63256

(I'm going to merge this as-is just to get it in)

PR #191, opened 2022-09-23
Test client on PR creation

beekley:
## Description

We now have client tests, so let's block merging if they don't pass

### Changed

- Adds the client test to the github workflow that runs when a PR is created.

## Testing and Reviewing

[It passes!](https://github.com/BlueConduit/open-data-platform/actions/runs/3114299609/jobs/5049980553)


justinebreuch (approved):
Nice! Should have thought of this

PR #193, opened 2022-09-23
Use geography types for area calculations

RP-3:
## Description

Addresses: [Fix tileserver function source](https://tables.area120.google.com/table/bX_kMNBW8LA35n6l5Hz_co/row/apv7QsIhjTr4ZTdhpDZ4lS)

- We should be casting polygons to `::geography` types for area calculations.
- See [here](http://postgis.net/workshops/postgis-intro/geography.html) for detailed explanation for why. 
- I'd failed to do this in schema.sql
- But this didn't affect prod and dev because I'd given folks the correct commands to run to manually fix it.
- This fix is live in production now. Nobody needs to do anything.
- If we need to make the same fix anywhere else, use the following:
```sql
UPDATE public.water_systems
SET approx_area_sq_km = ST_Area(ST_Transform(geom, 4326)::geography) / 1000000;
```

### New

- N/A

### Changed

- Area calculations now use a ::geography type so they return a meaningful area in sqKM.

### Removed

- N/A

## Testing and Reviewing

- The fix is live in prod. See the map to ensure it works: https://leadout.blueconduit.com/map

### DIFF
--- a/cdk/src/open-data-platform/data-plane/schema/schema.sql
+++ b/cdk/src/open-data-platform/data-plane/schema/schema.sql
@@ -11,7 +11,10 @@ CREATE EXTENSION IF NOT EXISTS postgis;
 --
 -- 4326 is a 3D coordinate system. Lat/Longs are implicitly converted to 3D space for the
 -- purposes of any calculations, including distances and areas. There is no way to display
--- 4326 coordinates in 2D without projecting it into 2D in some way or another.
+-- 4326 coordinates in 2D without projecting it into 2D in some way or another. The projection
+-- should either be to another spatial system (e.g., 3857 for tiles) or to another type that
+-- uses 'real' units for calculations, like ::geography.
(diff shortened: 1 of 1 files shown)


cdk/src/open-data-platform/data-plane/schema/schema.sql
```diff
@@ -11,7 +11,10 @@ CREATE EXTENSION IF NOT EXISTS postgis;
 --
 -- 4326 is a 3D coordinate system. Lat/Longs are implicitly converted to 3D space for the
 -- purposes of any calculations, including distances and areas. There is no way to display
--- 4326 coordinates in 2D without projecting it into 2D in some way or another.
+-- 4326 coordinates in 2D without projecting it into 2D in some way or another. The projection
```
justinebreuch, line 14: Thanks for adding the comment!

justinebreuch (approved):

PR #194, opened 2022-09-26
Refactor ZoomLevel to re-use GeographicLevel and hide pills in map view 

justinebreuch:
## Description

Addresses: [Hide pills on scorecard page and disable selection in side panel](https://tables.area120.google.com/u/1/workspace/bxQE2CPzwu_b4sJEkotkUM/table/bX_kMNBW8LA35n6l5Hz_co/view/auSM-s1cO9l3pAsYHmOk1E)

1. Hides the pills behind a feature called enableViewToggling in ScorecardSidePanel. Checked to make sure they work if turned on. 
2. Disables view selection behind enableSelection in GeoIdSelection
3. Removes "zoomLevel" from mapState in favor of geographicView that reuses GeographicLevel already referenced througout the UI. This caused issue because we had zoom and zoomLevel, different components were listening to different things causes discrepancies.

### Changed

ZoomLevel --> GeographicLevel 

### Removed

ZoomLevel

## Testing and Reviewing

Walked through all UI flows in Map and Scorecard. Logged new bugs found (also in dev).


client/src/model/store_plugin.ts
```diff
@@ -15,7 +15,6 @@ const createRedux = (store: EnhancedStore) => {
 
       store.subscribe(() => {
         rootStore.state = store.getState();
-        console.log(`The state has been updated ${JSON.stringify(rootStore.state)}`);
```
justinebreuch, line 18: This is very noisy

client/src/model/data_layer.ts
```diff
@@ -121,15 +121,17 @@ enum MapLayer {
 }
 
 /**
- * Zoom levels for each demographic layer.
+ * Zoom thresholds for each geographic level.
```
beekley, line 124: Does the enum applies if it's greater or less than the threshold? i.e. does `State` apply for 0 to 4, or 4 to 5?

(then could you add whatever the answer is to the comment here)

client/src/components/scorecard_page/PredictionPanel.vue
```diff
@@ -49,17 +51,21 @@ export default defineComponent({
...
-      if (this.showParcelPrediction)
+      }
+      if (this.showParcelPrediction) {
         return (
           this.formatPredictionAsLikelihood(this.publicLeadPercent) ??
           this.ScorecardSummaryMessages.NOT_ENOUGH_DATA_AVAILABLE
         );
-      if (this.showNoPrediction) return this.ScorecardSummaryMessages.NOT_ENOUGH_DATA_AVAILABLE;
+      }
+      if (this.showNoPrediction) {
```
beekley, line 66: (this is outside the scope of this PR) I don't love the negative for this property, when the others are positive. What's the difference between `showNoPrediction` and `!showParcelPrediction`?

client/src/data_layer_configs/population_by_census_block_config.ts
```diff
@@ -104,7 +104,7 @@ const legendInterpolation = [
 
 const legendInterpolationCounty = [
```
beekley, line 105: nit: Should this variable name change to say zip as well?

client/src/model/store_plugin.ts
beekley: +1. Though I'm up to still log some stuff that's useful.

beekley (approved):

client/src/model/data_layer.ts
justinebreuch: Added a comment to clarify thanks

client/src/data_layer_configs/population_by_census_block_config.ts
justinebreuch: Good catch

client/src/components/scorecard_page/PredictionPanel.vue
justinebreuch: showNoPrediction means there is neither parcel nor water system prediction

PR #195, opened 2022-09-26
Block pipeline on alarms

beekley:
## Description

Addresses: [Canaries](https://tables.area120.google.com/table/bX_kMNBW8LA35n6l5Hz_co/row/826xIGWGqdF62fj6Icp__4)

We have all these alarms, so this tells the pipeline to fail if there are any alarms firing.

### New

- After dev updates, the pipeline waits for 20 minutes then checks that there are no open alarms.
- Added a basic alarm for canary failures, plugged into slack.

### Changed

- Renamed the monitoring stacks to better match the other stacks. Before, it was naming them `Dev-Monitoring-Monitoring`. This *should* name them like `Dev-OpenDataPlatformMonitoring`, which will match the naming convention used by the pipeline.

## Testing and Reviewing

We won't really know until this hits the pipeline, since I can't locally test that. But I forced a canary failure and it fired the alarm (with [Slack notification](https://lslr.slack.com/archives/C03V1FX7KC1/p1664229316279529))

image

image


justinebreuch (commented):
LGTM but would wait for Aaron's review

kailamjeter (approved):

PR #205, opened 2022-09-29
Geoid in violations

kailamjeter:
Remove county from violation_counts view and add population of state census geo id to violations handler.

Tested in sandbox: https://kailajeter.leadout-sandbox.blueconduit.com/map?layer=epa-lead-and-copper-violations-by-water-system

### DIFF
--- a/cdk/src/open-data-platform/data-plane/data-import/write-violations-data.handler.ts
+++ b/cdk/src/open-data-platform/data-plane/data-import/write-violations-data.handler.ts
@@ -46,13 +46,18 @@ const insertIntoStatement = `INSERT INTO epa_violations (violation_id,
                                                          violation_code,
                                                          compliance_status,
                                                          start_date,
-                                                         end_date)
-                             VALUES (:violation_id,
-                                     :pws_id,
-                                     :violation_code,
-                                     :compliance_status,
-                                     :start_date,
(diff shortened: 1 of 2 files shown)


cdk/src/open-data-platform/data-plane/data-import/write-violations-data.handler.ts
```diff
@@ -46,13 +46,18 @@ const insertIntoStatement = `INSERT INTO epa_violations (violation_id,
...
-                                     :start_date,
-                                     :end_date) ON CONFLICT (violation_id) DO NOTHING`;
+                                                         end_date,
+                                                         state_census_geo_id)
+                             SELECT :violation_id,
+                                    :pws_id,
+                                    :violation_code,
+                                    :compliance_status,
+                                    TO_DATE(:start_date, 'YYYY-MM-DD'),
+                                    TO_DATE(:end_date, 'YYYY-MM-DD'),
```
justinebreuch, line 56: What happens here if end_date is null?

justinebreuch (approved):
Thank you!

cdk/src/open-data-platform/data-plane/data-import/write-violations-data.handler.ts
kailamjeter: never null now since I changed to default to empty string when null.

PR #210, opened 2022-10-03
Violations null date

kailamjeter:
Set end_date to null if it is empty. Without this the TO_DATE function returns a default date.

### DIFF
--- a/cdk/src/open-data-platform/data-plane/data-import/write-violations-data.handler.ts
+++ b/cdk/src/open-data-platform/data-plane/data-import/write-violations-data.handler.ts
@@ -53,8 +53,8 @@ const insertIntoStatement = `INSERT INTO epa_violations (violation_id,
                                     :violation_code,
                                     :compliance_status,
                                     TO_DATE(:start_date, 'YYYY-MM-DD'),
-                                    NULLIF(TO_DATE(:end_date, 'YYYY-MM-DD'),
-                                           TO_DATE('-0001-12-30')),
+                                    TO_DATE(NULLIF(:end_date, ''),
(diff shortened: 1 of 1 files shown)


cdk/src/open-data-platform/data-plane/data-import/write-violations-data.handler.ts
```diff
@@ -53,7 +53,8 @@ const insertIntoStatement = `INSERT INTO epa_violations (violation_id,
                                     :violation_code,
                                     :compliance_status,
                                     TO_DATE(:start_date, 'YYYY-MM-DD'),
-                                    TO_DATE(:end_date, 'YYYY-MM-DD'),
+                                    TO_DATE(NULLIF(:end_date, ''),
+                                            'YYYY-MM-DD'),
```
justinebreuch, line 57: Does this not give you errors? I had to do [this](https://github.com/BlueConduit/open-data-platform/blob/87202877551fd634b143b5aba6ceb3d687ba734e/cdk/src/open-data-platform/data-plane/model/parcels-table.ts#L104) for RDS to allow null writes

kailamjeter: No it doesn't. I think that error is caused by providing no value there (which you were able to resolve with isNull). We never let it get to that point here because a default empty string value is set on the builder. This checks if the value is still that default empty string and if so, sets the field to null, which is allowed because the field is nullable. 

I think this is essentially what you are doing, but your solution is a bit cleaner. I can switch this to that - didn't realize the isNull field existed.

PR #231, opened 2022-11-16
feat(cdk): make VPC creation env-independent

khalid-dev:
## Description

Addresses: [Deploy Leadout & Load Data in Dev Env](https://app.shortcut.com/blueconduit/story/6966/deploy-leadout-load-data-in-development-environment)

Currently we do a conditional lookup for an existing VPC resource when:
- deploying to `Development` via pipeline
- deploying to sandbox via `cdk deploy --all`

This lookup is done via a brittle hardcoded `vpcId`. Instead, this PR creates a fresh VPC in all deployment scenarios.

In the future, we may follow an internal pattern (established in `tributary`) to share the same VPC resource between dev sandboxes. This would be desirable because VPC resources are monetarily expensive - having one per dev is a bit cost inefficient. Holding off, though, until we have more clarity on development & maintenance expectations for this repo.

### New

- Ran `cdk synth --profile deployments` to add `Development` AZ lookup context to `cdk.context.json`

### Changed

- Creating VPC in every env instead of doing conditional lookup in sandbox & dev

### Removed

- Removed stale VPC lookup context from `cdk.context.json`

## Testing and Reviewing

- Clean up existing CloudFormation stacks in `Development` env <-- Wait to merge til this is done
- Merge PR & verify that pipeline successfully generates CloudFormation stacks in `Development` env

### DIFF
--- a/cdk/src/open-data-platform/network/network-stack.ts
+++ b/cdk/src/open-data-platform/network/network-stack.ts
@@ -4,7 +4,7 @@ import { Stack } from 'aws-cdk-lib';
 import * as ec2 from 'aws-cdk-lib/aws-ec2';
 import * as ecs from 'aws-cdk-lib/aws-ecs';
 import { Construct } from 'constructs';
-import { CommonProps, projectName, EnvType } from '../../util';
+import { CommonProps, EnvType } from '../../util';
 import { Dns } from './dns';
 
 export class NetworkStack extends Stack {
@@ -18,6 +18,7 @@ export class NetworkStack extends Stack {
     this.vpc = new ec2.Vpc(this, 'VPC', {
       cidr: '10.0.0.0/16',
       maxAzs: 2,
+      natGateways: props.envType !== EnvType.Production ? 1 : undefined,
       subnetConfiguration: [
         {
           cidrMask: 24,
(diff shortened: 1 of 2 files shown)


cdk/src/open-data-platform/network/network-stack.ts
```diff
@@ -15,44 +15,22 @@ export class NetworkStack extends Stack {
...
-            {
-              cidrMask: 20,
-              name: 'private',
-              subnetType: ec2.SubnetType.PRIVATE_WITH_NAT,
-            },
-          ],
-        });
+    this.vpc = new ec2.Vpc(this, 'VPC', {
+      cidr: '10.0.0.0/16',
+      maxAzs: 2,
```
arockett, line 20: Have maxAzs = 1 when not in Production. We only need the availability achieved by using mulitiple AZs for the env that users interact with. It's ok if Dev breaks because of an AWS AZ outage

arockett (changes requested):
Small change requested

cdk/src/open-data-platform/network/network-stack.ts
khalid-dev: Set nat gateways to `1` in non-prod envs 👍🏾 nice $catch ty

